PR #5, opened 2020-01-15
Function split block

nbouziani:
This branch has been derived from the `self_adjoint` branch (cf. #4). It contains an implementation of a split block that is working for the case where neither one of the split functions nor the function that was split is changed after the creation of the block (after the split). 

If it is the case, we still need a mechanism to take the new values into account in the dependency. Any ideas?

There is a failing example in `pyadjoint/tests/firedrake_adjoint/split_failing_example/` illustrating this problem.

@funsim  @sebastkm @dham @wence-


sebastkm:
The example in `split_block_failing.py` will not work because the subfunction is used as an initial guess for the nonlinear solver. We do not differentiate through the initial guess in pyadjoint, so even though it looks connected on the graph, that connection only works in the forward model.

But in general, to update `u` when updating `u1`, you want to use `FloatingType`. This is basically lazy annotation. So instead of initializing the blocks immediately, the subfunctions are informed of the blocks to initialize when they are updated.

See the Function.split implementation for fenics:
https://github.com/dolfin-adjoint/pyadjoint/blob/master/fenics_adjoint/types/function.py#L105

You can set the attributes directly after initialization instead of overloading `Function._split`.
Note that something similar to MergeBlock from fenics would solve the problem you describe, but the inverse could also happen: changes to `u` should also change `u1`, which is fixed by also having SplitBlock be lazy (i.e: using `block_class`)

PR #17, opened 2020-05-12
 `solve_adjoint` driver

jwallwork23:
Just a simple driver to solve the adjoint problem and endow the solve blocks with `adj_sol` values. In my work I’m not usually interested in computing gradients etc, only solving the adjoint problem. Doing it this way means we don’t need to choose an arbitrary control.

### DIFF
--- a/pyadjoint/drivers.py
+++ b/pyadjoint/drivers.py
@@ -69,3 +69,20 @@ def compute_hessian(J, m, m_dot, options=None, tape=None):
 
     r = [v.get_hessian(options=options) for v in m]
     return m.delist(r)
+
+
+def solve_adjoint(J, tape=None, adj_value=1.0):
+    """
+    Solve the adjoint problem for a functional J. The adjoint solutions are stored in the 'adj_sol'
+    attribute of each solve block.
+
+    Args:
+        J (AdjFloat):  The objective functional.
+        tape: The tape to use. Default is the current tape.
+    """
+    tape = tape or get_working_tape()
+    tape.reset_variables()
+    J.adj_value = adj_value
+
+    with stop_annotating():
+        tape.evaluate_adj(markings=False)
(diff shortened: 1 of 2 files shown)


pyadjoint/drivers.py
```diff
@@ -18,8 +18,8 @@ def compute_gradient(J, m, options=None, tape=None, adj_value=1.0):
         OverloadedType: The derivative with respect to the control. Should be an instance of the same type as
             the control.
     """
-    options = {} if options is None else options
-    tape = get_working_tape() if tape is None else tape
+    options = options or {}
```
stephankramer, line 21: I agree that this is cleaner - generally though, I'd be wary about "cleaning" code in places you aren't changing. It's likely to introduce new bugs.

jwallwork23:
Thanks for the feedback @stephankramer - good points. I'll modify the docstring.

pyadjoint/drivers.py
```diff
@@ -69,3 +69,22 @@ def compute_hessian(J, m, m_dot, options=None, tape=None):
 
     r = [v.get_hessian(options=options) for v in m]
     return m.delist(r)
+
+
+def solve_adjoint(J, tape=None, adj_value=1.0):
+    """
+    Solve the adjoint problem for a functional J.
+
+    Adjoint actions are stored on each tape block as the `adj_value` attribute. In the case of
```
stephankramer, line 78: Sorry, just to clarify: `adj_value`: is something that lives on a `BlockVariable` which represents a state of a Function/Constant/AdjFloat etc. not on blocks. The `adj_value` attribute on a block_variable accumulates the sensitivity of the functional wrt to that inbetween state when traversing backwards as you do here. the `adj_sol` attribute is something that lives on a solve block indeed. Maybe you want to say something like:

This traverses the entire tape backwards, unlike compute_gradient() which only works out those parts of the adjoint necessary to compute the sensitivity wrt the specified control. As a result sensitivities wrt to all inbetween states are accumulated in the adj_value attribute of the associated block-variables. The adjoint solution of each solution step is stored in the adj_sol attribute of the associated solve block.

jwallwork23: Thanks Stephan, have modified as suggested

PR #18, opened 2020-05-25
This fixes function assignment with expressions containing Constants

stephankramer:
Firedrake supports assignment with UFL expressions that can be evaluated
pointwise (in the same Functionspace). The annotation of these
previously failed to treat Constants as dependencies and compute the
relevant derivatives.


stephankramer:
Thanks! I believe so yes, hence only tested in firedrake_adjoint. More importantly it doesn't seem to break any existing fenics_adjoint functionality.

PR #21, opened 2020-06-25
Use equispaced coordinate function space in firedrake tests

tkarna:
Firedrake now uses equispaced function spaces for mesh coordinates which breaks firedrake adjoint tests. This PR fixes the function spaces definition in the tests.

### DIFF
--- a/tests/firedrake_adjoint/test_dynamic_meshes.py
+++ b/tests/firedrake_adjoint/test_dynamic_meshes.py
@@ -6,7 +6,7 @@ import numpy as np
 
 @pytest.mark.parametrize("mesh", [UnitSquareMesh(10,10)])
 def test_dynamic_meshes_2D(mesh):
-    S = VectorFunctionSpace(mesh, "CG", 1)
+    S = mesh.coordinates.function_space()
     s = [Function(S), Function(S), Function(S)]
     mesh.coordinates.assign(mesh.coordinates + s[0])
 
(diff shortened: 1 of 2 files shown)


tests/firedrake_adjoint/test_dynamic_meshes.py
```diff
@@ -1,12 +1,14 @@
 import pytest
 from firedrake import *
 from firedrake_adjoint import *
+import ufl
```
stephankramer, line 4: `FiniteElement` is already imported through firedrake.

tests/firedrake_adjoint/test_dynamic_meshes.py
```diff
@@ -1,12 +1,14 @@
...
 import numpy as np
 
 
 @pytest.mark.parametrize("mesh", [UnitSquareMesh(10,10)])
 def test_dynamic_meshes_2D(mesh):
-    S = VectorFunctionSpace(mesh, "CG", 1)
+    element = ufl.FiniteElement("Lagrange", mesh.ufl_cell(), 1, variant="equispaced")
+    S = VectorFunctionSpace(mesh, element)
```
stephankramer, line 11: Alternatively, I think you could do `S=mesh.coordinates.function_space()` ?

tkarna: ah yes, that's better because it would always just work

tests/firedrake_adjoint/test_dynamic_meshes.py
tkarna: fixed now

tests/firedrake_adjoint/test_dynamic_meshes.py
tkarna: fixed now

tkarna:
The tests pass with current firedrake version as well so it should be safe to merge this.

PR #37, opened 2020-10-30
tape: fixed tensorflow 2.0 incompatibility in visualise()

DiffeoInvariant:
Currently, `Tape.visualise()` does not work with TensorFlow 2.0 because TensorFlow 2.0 does not contain the `tf.Session` and `tf.summary.fileWriter` classes --- those are instead contained in `tf.compat.v1`. I've fixed this, so that `Tape.visualise()` works with TensorFlow 2.0 now.


sebastkm (approved):
Thanks!

PR #40, opened 2020-11-13
Add overloaded method for Function.sub

finsberg:
This is a fix for https://github.com/dolfin-adjoint/pyadjoint/issues/39

### DIFF
--- a/fenics_adjoint/types/function.py
+++ b/fenics_adjoint/types/function.py
@@ -58,8 +58,8 @@ class Function(FloatingType, backend.Function):
 
         return func
 
-    def sub(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


sebastkm:
This will not annotate the sub-function. To fix it we must make a sub-function that does the same as `dolfin_adjoint.Function.split`, but only for a single subspace instead of all of them. It should be relatively easy to implement. Let me know if you want to do it, or I can do it.

finsberg:
Great! Thanks for the guiding. I have modified the implementation to use `dolfin_adjoint.Function.split´ method.

sebastkm (approved):
Thanks! It would perhaps be better to test the annotation as well with a taylor test, but this is fine.

PR #50, opened 2021-02-24
Configure bibtex_bibfiles in conf.py

funsim:


dham (approved):
`conf.py` looks fine. I'm not sure why all this whitespace is going into tube_shape_derivative.rst

Issue #44, opened 2021-01-06
UFL constraints not available for Firedrake


tlroy:
UFL constraints are only available in fenics_adjoint/ufl_constraints. The example examples/stokes-topology/stokes-topology-rol-firedrake.py does not work as is. It's probably a simple fix to move ufl_constraints.py to pyadjoint instead.

I got it working for Firedrake by removing imports from fenics and firedrake_adjoint.types, which does not seem to exist.

PR #51, opened 2021-02-27
Incorporate UFLConstraints into Firedrake

medinaeder:
This PR should resolved issue #44 

The code has not been tested in the fenics environment. If someone could please verify that this does not break any functionality on that side that would be great.


sebastkm:
Thanks for the pull request.

The intention with pyadjoint is to be a generic framework for AD, and so it does not have a concept of a `backend` or UFL.
Moving `UFLConstraint` to `dolfin_adjoint_common` would be more suitable as this is where code that is shared between the fenics and firedrake implementations live.

medinaeder:
Sounds good. I have relocated the files.
Still seems to work.

sebastkm:
Nice. Could you also remove the UFL constraints file from `fenics_adjoint` and add a similar `__init__.py` import for `fenics_adjoint` using `dolfin_adjoint_common`?

sebastkm:
This looks good to me now.

medinaeder:
Great! Thanks for cleaning it up.

sebastkm (approved):

PR #57, opened 2021-03-11
Handle checkpoint correctly in minimize()

gbruer15:
There's a bug in `minimize_scipy_generic()` where it calls `ReducedFunctionalNumPy.obj_to_array()` on the output of `Control.data()`. 

```python
m = [p.data() for p in rf_np.controls]
m_global = rf_np.obj_to_array(m)
```

This is an issue because `.data()` returns the block_variable's checkpoint, which is not necessarily an OverloadedType, so to actually use it in `.obj_to_array()`, the checkpoint needs to be restored first.

I fixed this by calling `._ad_restore_at_checkpoint()`.

```python
m = [p._ad_restore_at_checkpoint(p.data()) for p in rf_np.controls]
```

Right now, this can't be tested easily, since all the OverloadedType implementations use their own type for the checkpoint. Let me know if there's a good way to test this that I should add.

### DIFF
--- a/pyadjoint/adjfloat.py
+++ b/pyadjoint/adjfloat.py
@@ -115,7 +115,7 @@ class AdjFloat(OverloadedType, float):
 
     @staticmethod
     def _ad_assign_numpy(dst, src, offset):
-        dst = float(src[offset:offset + 1])
+        dst = type(dst)(src[offset:offset + 1])
         offset += 1
         return dst, offset
 
--- a/pyadjoint/optimization/optimization.py
+++ b/pyadjoint/optimization/optimization.py
@@ -59,7 +59,7 @@ def minimize_scipy_generic(rf_np, method, bounds=None, **kwargs):
 
     project = kwargs.pop("project", False)
 
-    m = [p._ad_restore_at_checkpoint(p.data()) for p in rf_np.controls]
+    m = [p.tape_value() for p in rf_np.controls]
     m_global = rf_np.obj_to_array(m)
     J = rf_np.__call__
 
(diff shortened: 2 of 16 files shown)


sebastkm:
Good catch. It seems this assumption has been used for many other places in the code as well.
Perhaps we should move away from `Control.data` for most of these cases and use `Control.tape_value` instead, which essentially does the same as in your solution.

gbruer15:
That makes sense. I'll update it to use `Control.tape_value` and see where else it should be used.

gbruer15:
One thing that I wasn't quite sure about was in `pyadjoint/optimization/optimization_problem.py`. Should that allow both the checkpoint type and the overloaded type?

I also removed `Control.data` being used at the end of the minimization routines in optimization.py because `ReducedFunctionNumPy.set_controls` isn't guaranteed to change any of the tape values. It should instead be using the return value of `ReducedFunctionNumPy.set_controls`. An alternative solution would be to make that method modify the tape values by using `Control.update`.

I.e, it may fulfill the method's intention better if it was changed from this:
```python
    def set_controls(self, array):
        m = [p.tape_value() for p in self.controls]
        return self.set_local(m, array)
```
to this:
```python
    def set_controls(self, array):
        m = [p.tape_value() for p in self.controls]
        m = self.set_local(m, array)
        for c,v in zip(self.controls, m):
            c.update(v)
        return m
```

sebastkm:
Since the input to `OptimizationProblem` is from a user it should be the overloaded type and not checkpoint type. The original type (not overloaded) might also be acceptable though, although it should be rarely seen as users should only really be working with overloaded objects.

Aha, so the `set_controls` method assumes that the objects are mutable. I agree that the code snippet you provide would be more appropriate for the method's intention.
I see that `AdjFloat._ad_assign_numpy` actually returns a float regardless of the input type. At least it should return an overloaded type, but it is perhaps even better if it returns the same type as the input type (`dst`).

gbruer15:
Thanks for your help in reviewing this, Sebastian. I believe I've addressed the points you made. Let me know if you see anything else to improve here.

sebastkm:
This looks good to me now, I will merge later today if there are no further comments.
Thanks, Grant!

sebastkm (approved):

PR #58, opened 2021-03-11
Make create_overloaded_object() warning report the calling function

gbruer15:
This merge request just changes the warning in `create_overloaded_object()` to make the output much more useful.

Here's some example code to show how this looks:
```python
from pyadjoint import create_overloaded_object
create_overloaded_object('string')
```

Before this MR, the output was this:
```
$> python test.py 
/path/to/the/pyadjoint/repo/pyadjoint/overloaded_type.py:37: UserWarning: Could not find overloaded class of type '<class 'str'>'.
  warnings.warn("Could not find overloaded class of type '{}'.".format(obj_type))
```

This MR makes the output be this:
```
$> python test.py 
test.py:2: UserWarning: Could not find overloaded class of type '<class 'str'>'.
  create_overloaded_object('string')
```

### DIFF
--- a/pyadjoint/overloaded_type.py
+++ b/pyadjoint/overloaded_type.py
@@ -34,7 +34,7 @@ def create_overloaded_object(obj, suppress_warning=False):
     else:
         if not suppress_warning:
             import warnings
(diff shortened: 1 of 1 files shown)


sebastkm (approved):
Nice!

PR #63, opened 2021-05-05
Minor tweaks for AdjFloat

jwallwork23:
Two minor tweaks:
* use `cos` in the overloaded function for `CosBlock`, rather than `sin`;
* allow conversion from `AdjFloat` to `Constant`.

Maybe there is a nicer way to do the second point, but I didn't want to have to import the backend to check `isinstance(value, Constant)`.


jwallwork23:
Sorry, I realised this is covered by #47 . Never mind!

PR #66, opened 2021-06-15
Block tagging

jwallwork23:
I find this useful for sorting between different solve blocks written to tape when solving systems of equations.

### DIFF
--- a/dolfin_adjoint_common/blocks/assembly.py
+++ b/dolfin_adjoint_common/blocks/assembly.py
@@ -3,8 +3,8 @@ from pyadjoint import Block, create_overloaded_object
 
 
 class AssembleBlock(Block):
-    def __init__(self, form):
-        super(AssembleBlock, self).__init__()
+    def __init__(self, form, ad_block_tag=None):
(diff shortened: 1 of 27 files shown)


sebastkm:
Making it easier to differentiate between solves would be nice, indeed.

I wonder if perhaps it would be better with a more general pyadjoint approach, perhaps an optional argument to all annotated functions such as `ad_block_tag` which populates some `Block.tag` attribute.
Assuming the `options_prefix` attribute is not going to be used for anything inside the solve blocks.
This approach could also be complimented with some tape functions that return a list of blocks with a specific tag.

jwallwork23:
Good idea! I'll have a go at implementing something along those lines.

jwallwork23:
See also [Firedrake PR](https://github.com/firedrakeproject/firedrake/pull/2174).

jwallwork23:
@sebastkm What do you think of this approach? The Firedrake tests that fail here pass on my Firedrake PR (which tests against the pyadjoint branch here).

jwallwork23:
Thanks! Okay, great. I'll get someone to look at the corresponding Firedrake PR and then get them both merged together.

PR #78, opened 2022-03-16
Teach ROL Objective about update flags

angus-g:
ROL 2.0 calls the objective update method with a flag describing why the call was taken. This allows us to implement a caching scheme, to accept or revert updates if necessary. Ultimately, this can save quite a few objective function evaluations.

Because the ROL 2.0 update is very widely split, this change is backwards compatible, and will only be effective if pyrol wraps the UpdateType enumeration (e.g. https://github.com/angus-g/pyrol/commit/1c192fa6d8ffd0ae7f2648f9e38728f6ea546feb)

### DIFF
--- a/pyadjoint/optimization/rol_solver.py
+++ b/pyadjoint/optimization/rol_solver.py
@@ -1,5 +1,3 @@
-from __future__ import print_function
-
 from .optimization_solver import OptimizationSolver
 from ..enlisting import Enlist
 from ..overloaded_type import OverloadedType
@@ -14,9 +12,11 @@ try:
             self.rf = rf
             self.scale = scale
 
+            self._val = None
+            self._cache = None
+
         def value(self, x, tol):
-            # FIXME: should check if we have evaluated here before
-            return self.val
+            return self._val
 
         def gradient(self, g, x, tol):
             # self.rf(x.dat)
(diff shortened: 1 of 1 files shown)


sebastkm:
Very nice!

Funny that an update to pytest makes only the ROL tests fail, making it seem like this PR breaks something. I've merged a fix for that now.

Anyway, I wonder if this will work properly. I'm not familiar with the workings of ROL, but if say you store a `val` in `cache` and then trial a new `val` by `self.rf(x.dat)`, then if you revert back to the cached `val` and try to immediately afterwards call `gradient` the actually computed gradient will be at the rejected trial and not at the cached point (the gradient will be corresponding to the last call to `self.rf`).

Perhaps this will never happen with ROL, or perhaps a check needs to be also done during the gradient/hessian calls?

angus-g:
Hmm, that is a good point, there is indeed an example of exactly the sequence of events you described:
```cpp
    if (accept) obj.update(x,UpdateType::Accept,state_->iter);
    else        obj.update(x,UpdateType::Revert,state_->iter);
    obj.gradient(*state_->gradientVec,x,tol);
```

So I guess the only case when `rf` isn't re-evaluated is an accepted update.

PR #83, opened 2022-06-06
Convert ROLObjective to handle update flag

angus-g:
ROL 2.0 calls the objective update method with a flag describing why the call was taken. This allows us to implement a caching scheme, to accept or revert updates if necessary. Ultimately, this can save quite a few objective function evaluations.

Because the ROL 2.0 update is very widely split, this change is backwards compatible, and will only be effective if pyrol wraps the UpdateType enumeration (e.g. https://github.com/angus-g/pyrol/commit/1c192fa6d8ffd0ae7f2648f9e38728f6ea546feb)

This supersedes #78. I did expect the changes to be incompatible, but they ended up fairly similar. I've added some extra machinery to the tape to give us a map of block variables to their checkpointed values, which allows us to restore the tape to a consistent state that lets us re-evaluate the gradient correctly after a rejected trial.

### DIFF
--- a/pyadjoint/tape.py
+++ b/pyadjoint/tape.py
@@ -4,6 +4,7 @@ import re
 import threading
 from contextlib import contextmanager
 from functools import wraps
+from itertools import chain
 
 _working_tape = None
 _stop_annotating = 0
@@ -192,6 +193,18 @@ class Tape(object):
         """
         # TODO: Offer deepcopying. But is it feasible memory wise to copy all checkpoints?
         return Tape(blocks=self._blocks[:])
+
+    def block_vars(self, tag=None):
+        """Return a dictionary mapping all block variables on the tape to their checkpointed values"""
+
+        return {
+            var: var.checkpoint
+            for var in chain.from_iterable(
+                chain(
+                    b.get_dependencies(),
+                    b.get_outputs()
+                ) for b in self.get_blocks(tag))
+        }
 
     def optimize(self, controls=None, functionals=None):
         if controls is not None:
--- a/pyadjoint/tape.py
+++ b/pyadjoint/tape.py
@@ -194,16 +194,16 @@ class Tape(object):
         # TODO: Offer deepcopying. But is it feasible memory wise to copy all checkpoints?
         return Tape(blocks=self._blocks[:])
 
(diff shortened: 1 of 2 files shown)


dham (changes requested):
I think we need to encapsulate this a bit more, because with disk checkpointing there will be global tape state to save and restore as well. I think this would avoid the ROL solver having to access `_checkpoint` explicitly. I guess I am suggesting that we give the tape effectively checkpoint and restore methods. The first would return an opaque object which is the tape state, and the second would restore it. For current purposes, the object could be the dictionary you are currently using. Disk checkpointing would augment this dict with some global information (i.e. the checkpoint file locations).

angus-g:
Sounds reasonable to me, thanks!

pyadjoint/tape.py
```diff
@@ -193,6 +194,18 @@ def copy(self):
...
 
+    def block_vars(self, tag=None):
+        """Return a dictionary mapping all block variables on the tape to their checkpointed values"""
+
+        return {
+            var: var.checkpoint
+            for var in chain.from_iterable(
+                chain(
+                    b.get_dependencies(),
+                    b.get_outputs()
```
dham, line 205: I think this should only capture outputs. The reason is that anything which is only a dependency and not an output is actually an input to the tape. Its checkpoint should be left in place to be used on the next tape evaluation (and can safely be left on the tape because it won't be overwritten).

dham (changes requested):

pyadjoint/tape.py
angus-g: I think the problem with that as written is that there is at least one dependency to update: the control variable. I don't know of any way to update that explicitly, since it would normally only be set when evaluating the ReducedFunctional.

dham: True, but that doesn't mean we get to throw out all the other inputs. I think your argument means that we need all outputs and inputs which are also controls (the tape has a mechanism for marking controls).

angus-g: Makes sense! I've changed the method to accept a list of controls, from which we extract the `.block_variable` (like elsewhere in the tape methods).

dham (approved):

PR #86, opened 2022-06-15
Firedrake disk checkpoints

dham:
This PR primarily adds the facility for packages that use pyadjoint to attach their own state information to the tape and have it respond appropriately to tape events such as copy, clear, and reset. 

This is done to enable Firedrake disk checkpointing so there is also a small change to the `firedrake_adjoint` module and an additional test.

### DIFF
--- a/pyadjoint/overloaded_type.py
+++ b/pyadjoint/overloaded_type.py
@@ -333,7 +333,7 @@ class FloatingType(OverloadedType):
         if self._ad_floating_active:
             with FloatingType.stop_floating(self):
                 self._ad_annotate_block()
-        self.block_variable.save_output(overwrite=True)
+        self.block_variable.save_output(overwrite=False)
 
     def _ad_will_add_as_output(self):
         if self._ad_floating_active:
--- a/firedrake_adjoint/__init__.py
+++ b/firedrake_adjoint/__init__.py
@@ -14,6 +14,11 @@ from pyadjoint.tape import (Tape, set_working_tape, get_working_tape,
                             pause_annotation, continue_annotation,
                             stop_annotating, annotate_tape)
 from pyadjoint.reduced_functional import ReducedFunctional
+from firedrake.adjoint.checkpointing import (
+    enable_disk_checkpointing, pause_disk_checkpointing,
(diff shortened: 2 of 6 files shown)


pyadjoint/overloaded_type.py
```diff
@@ -333,7 +333,7 @@ def _ad_will_add_as_dependency(self):
         if self._ad_floating_active:
             with FloatingType.stop_floating(self):
                 self._ad_annotate_block()
-        self.block_variable.save_output(overwrite=True)
+        self.block_variable.save_output(overwrite=False)
```
dham, line 336: For the record, I've discussed this change with @sebastkm and he's fine with it.

Even floating types should only write to block variables once.

dham: I don't think we can detect this situation automatically. For starters, a dat can be shared by two Firedrake Functions. More generally, attempting to detect all of the circumstances in which a mutable overloaded type gets modified is impossible. 

What I think we could do is provide a front-door mechanism for the user to say that they are going to change the value of a variable that the adjoint knows about. Since we don't know what the user is going to do, we can only assume that the new value is unconnected to the old value and, as pointed out above, we achieve this by calling `.create_block_variable()`. 

To avoid any other issues that might occur due to taping what the user does while modifying the variable, annotation should be off. It seems to me that the nice way to do this would be to add an optional `modifies` argument to the `stop_annotating` context manager. This would take an OverloadedType or a sequence of such objects and, at the end of the context manager, would call `.create_block_variable`.  The last case above would then become:
```python3
u.assign(1.)
v1.project(u)
with stop_annotating(modifies=u):
    u.dat.data[:] = 2.
v2.project(u)
```
I will add this to the PR.

The particular case of the `dat.data` attribute is an important trip hazard in the case of Firedrake. I think that once this lands, we could spike that property so that it issues a warning if ever accessed with annotation switched on. I don't think there is a circumstance in which it would be necessary to have annotation switched on when accessing `.dat.data`. In the worst case one could just wrap that access in a stop_annotating context manager.

stephankramer (approved):
Thanks, all looks good to me. Added functionality to `stop_annotating` makes sense to me

PR #92, opened 2022-10-19
Extend AssembleBlock to handle 1-forms

nbouziani:
This PR augments the implementation of the TLM/Adjoint/Hessian of the AssembleBlock to cover 1-forms. Here is the associated Firedrake PR: https://github.com/firedrakeproject/firedrake/pull/2592

### DIFF
--- a/dolfin_adjoint_common/blocks/assembly.py
+++ b/dolfin_adjoint_common/blocks/assembly.py
@@ -19,6 +19,15 @@ class AssembleBlock(Block):
         return f"assemble({ufl2unicode(self.form)})"
 
     def compute_action_adjoint(self, adj_input, arity_form, form=None, c_rep=None, space=None, dform=None):
+        """This computes the action of the adjoint of the derivative of `form` wrt `c_rep` on `adj_input`.
+           In other words, it returns: `<(dform/dc_rep)*, adj_input>`
+
+           - If `form` has arity 0 => `dform/dc_rep` is a 1-form and `adj_input` a foat, we can simply use
+             the `*` operator.
+
+           - If `form` has arity 1 => `dform/dc_rep` is a 2-form and we can symbolically take its adjoint
+             and then apply the action on `adj_input`, to finally assemble the result.
+        """
         if arity_form == 0:
             if dform is None:
                 dc = self.backend.TestFunction(space)
@@ -42,13 +51,11 @@ class AssembleBlock(Block):
             else:
                 # Get PETSc matrix
                 dform_mat = self.compat.assemble_adjoint_value(dform).petscmat
-                # Adjoint (hermitian transpose)
-                dform_mat.hermitianTranspose()
-                # Action
+                # Action of the adjoint (Hermitian transpose)
                 adj_output = self.backend.Function(space)
                 with adj_input.dat.vec_ro as v_vec:
                     with adj_output.dat.vec as res_vec:
-                        dform_mat.mult(v_vec, res_vec)
+                        dform_mat.multHermitian(v_vec, res_vec)
             return adj_output, dform
         else:
             raise ValueError('Forms with arity > 1 are not handled yet!')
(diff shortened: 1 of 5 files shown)


dolfin_adjoint_common/blocks/assembly.py
```diff
@@ -18,6 +18,41 @@ def __init__(self, form, ad_block_tag=None):
...
+                # Get PETSc matrix
+                dform_mat = self.compat.assemble_adjoint_value(dform).petscmat
+                # Adjoint (hermitian transpose)
+                dform_mat.hermitianTranspose()
+                # Action
+                adj_output = self.backend.Function(space)
+                with adj_input.dat.vec_ro as v_vec:
+                    with adj_output.dat.vec as res_vec:
+                        dform_mat.mult(v_vec, res_vec)
+            return adj_output, dform
```
nbouziani, line 52: It would be good to have confirmation of this workaround. In terms of the implementation, the SpatialCoordinate case (l. 44-51) should simplify with the upcoming dual space changes in UFL which will make that case as the normal case (cf. line 40-41) except that we will call `Adjoint/Action` instead of `adjoint/action`.

Unlike to the 0-form case, the adjoint of 1-forms requires doing symbolic manipulation with the Jacobian. UFL doesn't support expanding coordinate derivatives of `Coefficient`s in physical space which leads to assembling the 2-form instead of symbolically treating it to only assemble a 1-form at the end (cf. line 40-41). I am not fully happy with this but I don't see another solution at the moment.

dolfin_adjoint_common/blocks/assembly.py
```diff
@@ -18,6 +18,41 @@ def __init__(self, form, ad_block_tag=None):
...
+            else:
+                # Get PETSc matrix
+                dform_mat = self.compat.assemble_adjoint_value(dform).petscmat
+                # Adjoint (hermitian transpose)
+                dform_mat.hermitianTranspose()
+                # Action
+                adj_output = self.backend.Function(space)
+                with adj_input.dat.vec_ro as v_vec:
+                    with adj_output.dat.vec as res_vec:
+                        dform_mat.mult(v_vec, res_vec)
```
dham, line 51: Delete line 46 and use `.multHermitian` here instead. This avoids having to actually make an adjoint matrix.

dolfin_adjoint_common/blocks/assembly.py
```diff
@@ -18,6 +18,41 @@ def __init__(self, form, ad_block_tag=None):
     def __str__(self):
         return f"assemble({ufl2unicode(self.form)})"
 
+    def compute_action_adjoint(self, adj_input, arity_form, form=None, c_rep=None, space=None, dform=None):
```
dham, line 21: Can we have a docstring please? I know we don't always but this one is at the complicated end so future poor sods who have to read this will thank us for at least saying what this does and what the arguments mean.

dolfin_adjoint_common/blocks/assembly.py
dham: I think this will actually work. I agree it's a bit disgusting but let's not get distracted by this.

Can you put in a test that's got a space derivative?

dham (changes requested):

dolfin_adjoint_common/blocks/assembly.py
nbouziani: Space derivatives are systematically tested for any assembled forms as they always turn up in the dependency. I have updated the initial control value of `test_assemble_1_forms_adjoint` to make that derivative bigger and therefore more sensitive to testing.

dolfin_adjoint_common/blocks/assembly.py
nbouziani: Thanks!

nbouziani:
@dham this PR is ready to be merged. The corresponding Firedrake PR has already been merged.

dham (approved):

PR #99, opened 2023-05-31
Reduced functional parameters

colinjcotter:
Provide an option to only compute a subset of the derivative components, returning zero in the other components.

use case: when you want to solve the same optimisation problem with different parameters, we need a way of resetting them. This can be fixed by adding them to the Controls. Then you can update their values via the __call__ method of ReducedFunctional.
However, then the optimisation algorithm will also optimise over those parameters as well, which we don't want. So the additional derivative_components argument to ReducedFunctional expresses that we only want to calculate derivatives in certain components of the controls (the components that are not parameters that we fix for each optimisation problem), and return zero for the other components. Then the gradient provided to the optimisation algorithm will be zero in those components, and it will not change the controls in those components during optimisation.

### DIFF
--- a/tests/pyadjoint/test_floats.py
+++ b/tests/pyadjoint/test_floats.py
@@ -256,8 +256,8 @@ def test_float_components():
     assert rf((AdjFloat(1.0), AdjFloat(2.0))) == 2.0
     assert rf((AdjFloat(3.0), AdjFloat(2.0))) == 6.0
     assert rf.derivative() == [0.0, 3.0]
-    assert rf((AdjFloat(3.0), AdjFloat(3.0))) == 9.0
-    assert rf.derivative() == [0.0, 3.0]
+    assert rf((AdjFloat(4.0), AdjFloat(3.0))) == 12.0
+    assert rf.derivative() == [0.0, 4.0]
 
 def test_float_components_minimize():
     a = AdjFloat(3.0)


pyadjoint/reduced_functional.py
```diff
@@ -4,6 +4,36 @@
 from .overloaded_type import OverloadedType
 
 
+def get_extract_derivative_components(derivative_components):
```
dham, line 7:
```suggestion
def _get_extract_derivative_components(derivative_components):
```

Please make these names private.

pyadjoint/reduced_functional.py
```diff
@@ -16,23 +46,35 @@ class ReducedFunctional(object):
...
+        derivative_components (tuple of int): The indices of the controls with
+            respect to which to take the derivative. By default, the derivative
+            is taken with respect to all controls. If present, it overwrites
+            derivative_cb_pre and derivative_cb_post.
+        derivative_cb_pre (function): Callback function before evaluating
+            derivatives. Should return a list of Controls (usually the same
+            list as the input) to be passed to compute_gradient.
+        derivative_cb_post (function): Callback function after evaluating
+            derivatives. Should return a list of Controls (usually the same
+            list as the input) to be passed to copmute_gradient.
```
dham, line 60: Is this really the interface for the post callback?

pyadjoint/reduced_functional.py
```diff
@@ -41,42 +83,55 @@ def __init__(self, functional, controls, scale=1.0, tape=None,
...
         """
+
         # Call callback
         values = [c.tape_value() for c in self.controls]
-        self.derivative_cb_pre(self.controls.delist(values))
+        controls = self.derivative_cb_pre(self.controls)
+        if not isinstance(controls, Enlist):
+            raise DeprecationWarning("""Note that the callback interface
+            for derivative_cb_pre has changed. It should now return a
+            list of controls (usually the same list as input.""")
```
dham, line 116: Error message doesn't match if statement. Is it `list` or `Enlist`?

pyadjoint/reduced_functional.py
dham: What arguments to the callbacks take?

dham (changes requested):

pyadjoint/reduced_functional.py
colinjcotter: fixed in latest commit.

pyadjoint/reduced_functional.py
colinjcotter: fixed in latest commit.

pyadjoint/reduced_functional.py
colinjcotter: Fixed in latest commit - I just check that the callback returns something now, because we just need to alert people to the change in interface.

dham (approved):

PR #112, opened 2023-09-07
Sanitize adjoint input of ReducedFunctional

nbouziani:
Ensure that the adjoint input of `ReducedFunctional` is an overloaded type and make use of the corresponding scaling method instead of relying on Python `*`.

### DIFF
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -1,7 +1,7 @@
 from .drivers import compute_gradient, compute_hessian
 from .enlisting import Enlist
 from .tape import get_working_tape, stop_annotating, no_annotations
-from .overloaded_type import OverloadedType
+from .overloaded_type import OverloadedType, create_overloaded_object
 
 
 def _get_extract_derivative_components(derivative_components):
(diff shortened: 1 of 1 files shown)


pyadjoint/reduced_functional.py
```diff
@@ -123,13 +123,15 @@ def derivative(self, adj_input=1.0, options={}):
 
         # Scale adjoint input
         with stop_annotating():
-            adj_value = self.scale * adj_input
+            # Make sure `adj_input` is an OverloadedType
+            adj_input = create_overloaded_object(adj_input)
+            adj_input._ad_mul(self.scale)
```
jorgensd, line 128:
```suggestion
            adj_value = adj_input._ad_mul(self.scale)
```
This operation does not do in-place, it returns the product

pyadjoint/reduced_functional.py
```diff
@@ -123,13 +123,15 @@ def derivative(self, adj_input=1.0, options={}):
...
 
         derivatives = compute_gradient(self.functional,
                                        controls,
                                        options=options,
                                        tape=self.tape,
-                                       adj_value=adj_value)
+                                       adj_value=adj_input)
```
jorgensd, line 134:
```suggestion
                                       adj_value=adj_value)
```

jorgensd (approved):

PR #118, opened 2023-09-26
Update ROLObjective to be dualspace-aware

angus-g:
ROLObjective was deferring to ROLVector to use `_ad_convert_type` (through `ROLVector.riesz_map`) on the result of `rf.derivative()` or `rf.hessian()`. This gives a type error, where a Cofunction is expected. Because (as far as I can tell), the derivative/Hessian operations can return a Riesz representation according to a specific inner product, we should pass this through these calls, instead of a separate `_ad_convert_type` step.

Note that I haven't dealt with the calls to `ROLVector.riesz_map` from ROLConstraint, but perhaps the fix there is equally simple? Hopefully this is the right way to go about making the ROL interface dualspace-aware, anyway.

### DIFF
--- a/pyadjoint/optimization/rol_solver.py
+++ b/pyadjoint/optimization/rol_solver.py
@@ -20,8 +20,9 @@ try:
             return self._val
 
         def gradient(self, g, x, tol):
-            self.deriv = self.rf.derivative()
-            g.dat = g.riesz_map(self.deriv)
+            opts = {"riesz_representation": x.inner_product}
+            self.deriv = self.rf.derivative(options=opts)
+            g.dat = Enlist(self.deriv)
 
         def hessVec(self, hv, v, x, tol):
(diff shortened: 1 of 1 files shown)


sghelichkhani:
> Note that I haven't dealt with the calls to ROLVector.riesz_map from ROLConstraint, but perhaps the fix there is equally simple? Hopefully this is the right way to go about making the ROL interface dualspace-aware, anyway.
Does this mean that providing constraints with ROL would fail/not work properly?

angus-g:
> Does this mean that providing constraints with ROL would fail/not work properly?

Only if `applyAdjointJacobian` or `applyAdjointHessian` is called on the constraint, which doesn't happen for most of the "simple" type-B algorithms, except for Primal-Dual Active-Set (for which there isn't an interface at the moment).

dham (approved):

PR #121, opened 2023-10-19
Dolci/checkpointing time depend

Ig-dolci:


pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,305 @@
...
+    index = 0
+    forward_steps = 0
+
+    while not isinstance(schedule[index], EndForward):
+        if isinstance(schedule[index], Forward):
+            forward_steps += len(schedule[index])
+        index += 1
+    end_forward = index + 1
+    reverse_steps = 0
+
```
connorjward, line 59: Probably don't want whitespace here. This doesn't separate a logical chunk of code. Perhaps the line above?

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,305 @@
...
+    return forward, reverse
+
+
+class CheckpointManager:
+    """Manage the executions of the forward and adjoint solvers.
+
+    Attributes
+    ----------
+    schedule : checkpoint_schedules.schedule
+        A schedule provided by the checkpoint_schedules package.
```
connorjward, line 76: Should mention that only `Revolve` is currently supported. Otherwise users only discover this when they run their code.

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,305 @@
...
+        self._stored_timesteps = []
+        self._shedule = schedule
+        self._checkpointable_state = set()
+        # Tell the tape to only checkpoint input data until told otherwise.
+        self.tape.latest_checkpoint = 0
+        # Process any initial instructions on the tape.
+        self.end_timestep(-1)
+
+    def end_timestep(self, timestep):
+        """Process any initial instructions on the tape.
```
connorjward, line 106: I don't think that this docstring really matches the name of the function. It's not very clear what happens.

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,305 @@
...
+        """A single-dispatch generic function.
+        Process taping is used while the forward model is taped.
+
+        Parameters
+        ----------
+        cp_action : schedules.CheckpointAction
+            A checkpoint action from the schedule. The actions can
+            be `Forward`, `Reverse`, `EndForward`, `EndReverse`, `Copy`, or `Move`.
+        timestep : int
+            The current timestep.
```
connorjward, line 138: Should document `Returns`.

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,305 @@
...
+        """Process the end of the forward execution."""
+        current_timestep = self.tape.latest_timestep
+        while self.mode != Mode.EVALUATED:
+            self.end_timestep(current_timestep)
+            current_timestep += 1
+
+    @singledispatchmethod
+    def process_taping(self, cp_action, timestep):
+        """A single-dispatch generic function.
+        Process taping is used while the forward model is taped.
```
connorjward, line 130: Not imperative mood.

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,305 @@
...
+        last_block : int
+            The last block to be evaluated.
+        markings : bool
+            If True, then each block_variable of the current block will have
+            set `marked_in_path` attribute indicating whether their adjoint
+            components are relevant for computing the final target adjoint
+            values. Default is False.
+        """
+        # Work out other cases when they arise.
+        assert last_block == 0
```
connorjward, line 210: Should probably `raise NotImplementedError` or something. Assertion errors are just for internal code.

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,305 @@
...
+                to_keep = next_step.checkpointable_state
+                if functional:
+                    # to_keep holds the variables required for restarting
+                    # the forward model from a step `n`.
+                    to_keep = to_keep.union([functional.block_variable])
+                for block in current_step:
+                    for var in block.get_outputs():
+                        if var not in to_keep:
+                            var._checkpoint = None
+                    # This code is making the burger's test fail.
```
connorjward, line 252: Needs addressing.

pyadjoint/reduced_functional.py
```diff
@@ -206,10 +206,13 @@ def __call__(self, values):
         blocks = self.tape.get_blocks()
         with self.marked_controls():
             with stop_annotating():
-                for i in self.tape._bar("Evaluating functional").iter(
-                    range(len(blocks))
-                ):
-                    blocks[i].recompute()
+                if self.tape._checkpoint_manager is not None:
```
JDBetteridge, line 209:
```suggestion
                if self.tape._checkpoint_manager:
```

pyadjoint/tape.py
```diff
@@ -226,10 +308,25 @@ def get_tags(self):
...
-        for i in self._bar("Evaluating adjoint").iter(
-            range(len(self._blocks) - 1, last_block - 1, -1)
-        ):
-            self._blocks[i].evaluate_adj(markings=markings)
+        """Evaluate the adjoint of the tape.
+
+        Parameters
+        ----------
+        last_block : int, optional
+            The index of the last block to evaluate. Default 0.
```
connorjward, line 316:
```suggestion
            The index of the last block to evaluate.
```
this is obvious from the function definition.

pyadjoint/tape.py
```diff
@@ -226,10 +308,25 @@ def get_tags(self):
...
+        markings : bool, optional
+            If True, then each :class:`BlockVariable` of the current block
+            will have set `marked_in_path` attribute indicating whether their
+            adjoint components are relevant for computing the final target
+            adjoint values. Default is False.
```
connorjward, line 321:
```suggestion
            adjoint values.
```

pyadjoint/tape.py
```diff
@@ -226,10 +308,25 @@ def get_tags(self):
...
-        ):
-            self._blocks[i].evaluate_adj(markings=markings)
+        """Evaluate the adjoint of the tape.
+
+        Parameters
+        ----------
+        last_block : int, optional
+            The index of the last block to evaluate. Default 0.
+        markings : bool, optional
+            If True, then each :class:`BlockVariable` of the current block
```
connorjward, line 318:
```suggestion
            If True, then each `BlockVariable` of the current block
```
I think. You should check the autogenerated docs.

connorjward (dismissed):
Style comments.

pyadjoint/tape.py
```diff
@@ -160,11 +160,13 @@ class Tape(object):
...
     __slots__ = ["_blocks", "_tf_tensors", "_tf_added_blocks", "_nodes",
-                 "_tf_registered_blocks", "_bar", "_package_data"]
+                 "_tf_registered_blocks", "_bar", "_package_data",
+                 "_checkpoint_manager", "latest_checkpoint",
+                 "_eagerly_checkpoint_outputs"]
 
     def __init__(self, blocks=None, package_data=None):
         # Initialize the list of blocks on the tape.
-        self._blocks = [] if blocks is None else blocks
+        self._blocks = TimeStepSequence(blocks=blocks) if blocks is not None else TimeStepSequence()
```
JDBetteridge, line 169:
```suggestion
        self._blocks = TimeStepSequence(blocks=blocks) if blocks else TimeStepSequence()
```

pyadjoint/tape.py
```diff
@@ -200,6 +244,44 @@ def add_block(self, block):
...
+        forward and adjoint computations.
+
+        Parameters
+        ----------
+        schedule : checkpoint_schedules.schedule
+            A schedule provided by the checkpoint_schedules package.
+        max_n : int, optional
+            The number of total steps.
+        """
+        if len(self._blocks) > 0:
```
JDBetteridge, line 279:
```suggestion
        if self._blocks:
```

pyadjoint/tape.py
```diff
@@ -319,40 +416,59 @@ def optimize(self, controls=None, functionals=None):
...
+            optimized_timesteps.steps[-1].checkpointable_state = \
+                step.checkpointable_state - discarded_variables
+
+        self._blocks = optimized_timesteps
 
     def optimize_for_functionals(self, functionals):
-        blocks = self.get_blocks()
-        nodes = set([functional.block_variable for functional in functionals])
-        valid_blocks = []
+        retained_nodes = set([functional.block_variable
```
JDBetteridge, line 446: parallel safe?

pyadjoint/tape.py
```diff
@@ -596,6 +717,104 @@ def iter(self, iterator):
...
+        super().__init__(blocks)
+        # The set of block variables which are needed to restart from the start
+        # of this timestep.
+        self.checkpointable_state = set()
+        # A dictionary mapping the block variables in the checkpointable state
+        # to their checkpoint values.
+        self._checkpoint = {}
+
+    def copy(self, blocks=None):
+        out = TimeStep(self) if blocks is None else TimeStep(blocks)
```
JDBetteridge, line 794:
```suggestion
        out = TimeStep(self) if not blocks else TimeStep(blocks)
```

JDBetteridge:
```suggestion
        out = TimeStep(blocks) if blocks else TimeStep(self)
```
maybe better

JDBetteridge (dismissed):
Mainly style, some concerns about `set()` being unordered and how this will work in parallel, but I don't 100% follow all the logic here

pyadjoint/block_variable.py
```diff
@@ -60,12 +64,22 @@ def saved_output(self):
...
+            self.save_output(overwrite=overwrite)
+        tape.add_to_checkpointable_state(self, self.last_use)
+        self.last_use = tape.latest_timestep
 
     def will_add_as_output(self):
+        tape = get_working_tape()
+        self.creation_timestep = tape.latest_timestep
+        self.last_use = self.creation_timestep
         overwrite = self.output._ad_will_add_as_output()
         overwrite = True if overwrite is None else overwrite
```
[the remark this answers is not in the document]
connorjward: I don't think this suggestion is correct. The current system sort of inverts the "truthiness" of `None` (i.e. `overwrite == None` means `overwrite == True`) whereas your suggestion inverts the truthiness of bools too.

pyadjoint/reduced_functional.py
connorjward: I don't think that this is always the right thing to do. From [PEP8](https://peps.python.org/pep-0008/#programming-recommendations):
> Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,328 @@
...
+        ):
+            raise CheckpointError(
+                "The schedule employs disk checkpointing but it is not configured."
+            )
+        self.tape = tape
+        self._schedule = schedule
+        self.forward_schedule = []
+        self.reverse_schedule = []
+        self.timesteps = schedule.max_n
+        # This variable is used to indicate whether the adjoint model
```
jorgensd, line 59:
```suggestion
        # This variable is used to indicate whether the adjoint model has been evaluated at the checkpoint.
```

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,328 @@
...
+    EVALUATE_ADJOINT = 6
+
+
+class CheckpointManager:
+    """Manage the executions of the forward and adjoint solvers.
+
+    Attributes
+    ----------
+    schedule : checkpoint_schedules.schedule
+        A schedule provided by the checkpoint_schedules package.
```
jorgensd, line 28:
```suggestion
        A schedule provided by the `checkpoint_schedules` package.
```

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,328 @@
...
+        if timestep < (cp_action.n0):
+            raise CheckpointError(
+                "Timestep is before start of Forward action."
+            )
+
+        self.tape._eagerly_checkpoint_outputs = cp_action.write_adj_deps
+
+        if timestep > cp_action.n0:
+            if cp_action.write_ics and timestep == (cp_action.n0 + 1):
+                # Stores the checkpointint data in RAM
```
jorgensd, line 135:
```suggestion
                # Stores the checkpoint data in RAM
```

pyadjoint/tape.py
```diff
@@ -596,6 +718,104 @@ def iter(self, iterator):
         return iterator
 
 
+class TapeTimeStepper:
+    """Iterator wrapper which advances the timestep after each iteration."""
+    def __init__(self, tape, iterable):
+        self.tape = tape
+        self.iterator = tape._bar("Taping forward").iter(iterable)
```
jorgensd, line 725:
```suggestion
        self.iterator = tape.progress_bar("Taping forward").iter(iterable)
```

pyadjoint/tape.py
```diff
@@ -174,12 +176,54 @@ def __init__(self, blocks=None, package_data=None):
...
+        iterable : iterable
+            The iterable definining the sequence of timesteps.
+
+        Notes
+        -----
+        This method facilitates taping timestepping simulations so that recompute
+        checkpointing can be used on the tape. For example, a simulation with
+        10 timesteps might use a timestepping loop of this form::
+
+            tape = get_working_tape
```
jorgensd, line 219:
```suggestion
            tape = get_working_tape()
```

pyadjoint/tape.py
```diff
@@ -319,40 +417,59 @@ def optimize(self, controls=None, functionals=None):
...
+        optimized_timesteps = TimeStepSequence()
+
+        for step in self._blocks.steps:
+            optimized_timesteps.append_step()
+
+            for block in step:
+                depends_on_control = False
+                for dep in block.get_dependencies():
+                    if dep in nodes:
+                        depends_on_control = True
```
jorgensd, line 431:
```suggestion
                        depends_on_control = True
                        break
```
I would suggest adding a break here, as we shouldn't need to continue looping through dependencies if we have found a dependency in an earlier node/

pyadjoint/tape.py
```diff
@@ -160,11 +160,13 @@ class Tape(object):
...
     __slots__ = ["_blocks", "_tf_tensors", "_tf_added_blocks", "_nodes",
-                 "_tf_registered_blocks", "_bar", "_package_data"]
+                 "_tf_registered_blocks", "_bar", "_package_data",
+                 "_checkpoint_manager", "latest_checkpoint",
+                 "_eagerly_checkpoint_outputs"]
 
     def __init__(self, blocks=None, package_data=None):
         # Initialize the list of blocks on the tape.
-        self._blocks = [] if blocks is None else blocks
+        self._blocks = TimeStepSequence(blocks=blocks if blocks else ())
```
jorgensd, line 169:
```suggestion
        self._blocks = TimeStepSequence(blocks=blocks if blocks else [])
```
Feels wrong to send in an empty tuple, as they are immutable, and gives the idea that one cannot append blocks at a later stage.

pyadjoint/drivers.py
```diff
@@ -91,3 +92,16 @@ def solve_adjoint(J, tape=None, adj_value=1.0):
 
     with stop_annotating():
         tape.evaluate_adj(markings=False)
+
+
+class marked_controls:
```
jorgensd, line 97: Document class.
For instance,
"""
Context manager for an iterable of :class:`Control`.
At entry point the controls will all be marked (Made relevant for tape optimizations).
At exit of the context manager, the controls will be reset.
"""

To simplify it, we could just use `contextlib`: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,328 @@
...
+        Notes
+        -----
+        Additional details about checkpoint_schedules can be found at
+        checkpoint_schedules
+        `documentation <https://www.firedrakeproject.org/checkpoint_schedules/>`_.
+        """
+        raise CheckpointError(f"Unable to process {cp_action} while taping.")
+
+    @process_taping.register(Forward)
+    def _(self, cp_action, timestep):
```
jorgensd, line 125: It would be nice to have some documentation of what the `process_taping` does when entering forward mode, or at least have a link to:
https://www.firedrakeproject.org/checkpoint_schedules/checkpoint_schedules.html#checkpoint_schedules.schedule.Forward

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,328 @@
...
+            set `marked_in_path` attribute indicating whether their adjoint
+            components are relevant for computing the final target adjoint
+            values.
+        """
+        # Work out other cases when they arise.
+        if last_block != 0:
+            raise NotImplementedError(
+                "Only the first block can be evaluated at present."
+            )
+
```
jorgensd, line 204: Shouldn't this error message be somewhere else than inside `evalute_adj`, as `last_block` is never used inside this function, except for the error checking?

jorgensd (changes requested):
Looks good.
There are some comments regarding documentation and style in certain places.

Would also be good to a have a simple test for this in `pyadjoint`, not just in the `firedrake.adjoint`/`firedrake_adjoint` framework.

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,343 @@
...
+        self.tape._eagerly_checkpoint_outputs = cp_action.write_adj_deps
+
+        if timestep > cp_action.n0:
+            if cp_action.write_ics and timestep == (cp_action.n0 + 1):
+                # Stores the checkpoint data in RAM
+                # This data will be used to restart the forward model
+                # from the step `n0` in the reverse computations.
+                self.tape.timesteps[cp_action.n0].checkpoint()
```
jrmaddison, line 147: Can I check about the case were forward data computed in step $n$ is used for forward calculations in steps $\ge n + 2$.

e.g. for the Firedrake code
```
from firedrake import *
from firedrake.adjoint import continue_annotation

mesh = UnitIntervalMesh(1)
space = FunctionSpace(mesh, "DG", 0)
U = [Function(space).assign(n) for n in range(N)]

continue_annotation()
v = Function(space)
for u in get_working_tape().timestepper(U):
   v = Function(space).assign(v + u)
```
the elements of `U` all have `last_use = -1`, and are each used in exactly one step. A checkpoint for a forward restart at step the start of step `n0` advancing to the start of step `n1` needs to include `U[n0:n1]`.

Storing the checkpoint after advancing to the start of step `n0 + 1` looks wrong, unless later `U` elements are flagged for checkpointing at step `n0`. Including too many `U` elements uses too much storage.

I might just be missing how this works (I see other relevant code in tape.py), but I suggest adding a test case.

Ig-dolci:
Notes: Move the docstring from numpy-style to google-style.

pyadjoint/checkpointing.py
jrmaddison: A different example, with a structure that might be encountered with an explicit linear multistep discretization:

```
from firedrake import *
from firedrake.adjoint import continue_annotation, get_working_tape

from collections import deque

M, N = 3, 100

mesh = UnitIntervalMesh(1)
space = FunctionSpace(mesh, "DG", 0)
U = deque([Function(space).assign(m) for m in range(M)])

continue_annotation()
v = Function(space)
for _ in get_working_tape().timestepper(range(N)):
   U[0].assign(U[-1])
   U.append(U.popleft())
```

Ig-dolci:
> Looks good. There are some comments regarding documentation and style in certain places.
> 
> Would also be good to a have a simple test for this in `pyadjoint`, not just in the `firedrake.adjoint`/`firedrake_adjoint` framework.

It is hard to build a pyadjoint test since the checkpointing will not work with `AdjFloat`. I do not understand deeply, but I think be no possible build `Blocks` and its outputs and dependencies with test based in arrays.

pyadjoint/checkpointing.py
```diff
@@ -0,0 +1,305 @@
...
+    """
+    def __init__(self, cp_action=(), steps=0):
+        # Steps is only intended to provide a rough indicator for progress
+        # bars.
+        super().__init__(cp_action)
+        self.steps = steps
+
+
+def process_schedule(schedule):
+    """Process a checkpoint schedule into forward and adjoint schedules.
```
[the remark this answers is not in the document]
Ig-dolci: This function was removed.

pyadjoint/checkpointing.py
Ig-dolci: done.

jorgensd (approved):
Looks ok to me. I'll let @dham decide when he is happy with this.

dham (approved):

Issue #111, opened 2023-09-04
Unintuitive taping behaviour


connorjward:
I have always found it confusing that the enabling of taping is [handled using an integer](https://github.com/dolfin-adjoint/pyadjoint/blob/master/pyadjoint/tape.py#L12) rather than a boolean. It means that one can have what I believe to be very unintuitive behaviour such as the following:

```py
from pyadjoint.tape import continue_annotation, annotate_tape, stop_annotating

continue_annotation()
with stop_annotating():
    assert not annotate_tape()  # raises an error
```

What this MFE does is start annotation twice (one happens at import) and only stop it once. Which, since an integer is used, means that annotation is still considered as active inside the `stop_annotating` context manager.

Is this a bug or a feature?

dham:
This isn't my code, but I think that what is going on here is an attempt to make the following code work correctly:

```
from pyadjoint.tape import continue_annotation, annotate_tape, stop_annotating

pause_annotation()
with stop_annotating():
    pass
 assert not annotate_tape()  # raises an error
```

This explains why a plain boolean isn't sufficient. However, what is currently being done is abusing an integer as an incorrectly implemented stack. It would be better to have an actual stack. The easiest way to implement this would probably be to have the `stop_annotation` context manager keep the state of annotation before it executes and restore that.

connorjward:
> This explains why a plain boolean isn't sufficient. However, what is currently being done is abusing an integer as an incorrectly implemented stack. It would be better to have an actual stack. The easiest way to implement this would probably be to have the `stop_annotation` context manager keep the state of annotation before it executes and restore that.

I can see how something like

```py
class stop_annotating:
    def __enter__(self):
        global _annotation_enabled
        _annotation_enabled = False

    def __exit__(self, *args):
        global _annotation_enabled
        _annotation_enabled = True  # not always the right thing to do
```

would be wrong. Are you suggesting instead that everything would work if we had something like

```py
class stop_annotating:
    def __enter__(self):
        global _annotation_enabled
        self._orig_annotation_enabled = _annotation_enabled
        _annotation_enabled = False

    def __exit__(self, *args):
        global _annotation_enabled
        _annotation_enabled = self._orig_annotation_enabled
```

I just want to be clear that when you say "have an actual stack" you don't mean to use a list or similar to pop and push from. The state awareness of the context manager is sufficient.

dham:
[...]

PR #129, opened 2024-01-20
Update tape documentation

prjaga:
Documentation used to be 

> Note that if you call pause_annotation() twice, then continue_annotation() must be called twice to enable annotation. Due to this, the recommended annotation control functions are stop_annotating and no_annotations().

updated this to reflect changes in #111 to

> Annotation can be temporarily disabled using pause_annotation() and enabled again using continue_annotation(). It is recommended to use stop_annotating and no_annotations() for annotation control.

as there is no counter anymore there is no need to call these functions twice to re-enable annotating.


dham (approved):

dham:
Thank you for catching this @prjaga

PR #138, opened 2024-02-26
Remove hint that is not working for python3.8

Ig-dolci:
Fix Python 3.8 issue in `TimeStepSequence` related to the [hint](https://github.com/dolfin-adjoint/pyadjoint/blob/d752d79e4ed9bc39c4822475d8a79627050464a7/pyadjoint/tape.py#L775). This commit aims to address the problem by introducing a class docstring to provide information about the optional `TimeStepSequence` arguments. The issue appears to be related to the `Iterable` type, as demonstrated by the traceback below:

```bash
python3.8 -c "from collections.abc import Iterable; Iterable[list]"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: 'ABCMeta' object is not subscriptable
```

Adding the class docstring aims to enhance clarity and help fix the Python 3.8 compatibility issue.

### DIFF
--- a/pyadjoint/tape.py
+++ b/pyadjoint/tape.py
@@ -6,8 +6,6 @@ from contextlib import contextmanager
 from functools import wraps
 from itertools import chain
 from abc import ABC, abstractmethod
-from typing import Optional
-from collections.abc import Iterable
 from .checkpointing import CheckpointManager, CheckpointError
 
 _working_tape = None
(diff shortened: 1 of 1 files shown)


jorgensd:
Could you elaborate on what specific part of that type hint that doesn't work with Python 3.8?

Ig-dolci:
> Could you elaborate on what specific part of that type hint that doesn't work with Python 3.8?
Yes. See this code below:
```
python3.8 -c "from collections.abc import Iterable; Iterable[list]"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: 'ABCMeta' object is not subscriptable
```

connorjward:
Might a simpler fix be to replace the `from collections.abc import Iterable` with `from typing import Iterable`? The ability to use `collections.abc.Iterable[some_type]` [was only added in 3.9](https://docs.python.org/3/library/typing.html#typing.Iterable).

Ig-dolci:
> Might a simpler fix be to replace the `from collections.abc import Iterable` with `from typing import Iterable`? The ability to use `collections.abc.Iterable[some_type]` [was only added in 3.9](https://docs.python.org/3/library/typing.html#typing.Iterable).

Sounds good. It is working for python3.8 and the more recent versions. I think we can keep this docstring anyway.

pyadjoint/tape.py
```diff
@@ -770,6 +769,14 @@ class TimeStepSequence(list):
 
     This behaves like a list of blocks. To access a list of the timesteps, use
     the :attr:`steps` property.
+
+    Args:
+        blocks (list[Block] or TimeStepSequence, optional): If provided, `blocks` can be a list of :class:`Block`
+        or another TimeStepSequence to copy from.
+        steps (list[TimeStep], optional): If provided, `steps` should be a list of :class:`TimeStep` to copy from.
+
+    Atributes:
```
connorjward, line 778:
```suggestion
    Attributes:
```

PR #143, opened 2024-05-02
TAO interface

jrmaddison:
Optimisation using PETSc/TAO.

Uses the pyadjoint `OverloadedType` interface to provide a generic implementation. However these interfaces don't provide enough information about data decomposition with MPI parallelism. This is worked around in this PR using `OverloadedType._ad_assign_numpy`, but this requires a temporary, but global, NumPy array. Fixing this requires a change to `OverloadedType` and then other updates elsewhere (in particular in Firedrake).

Related, `OverloadedType._ad_to_list` returns a `list` rather than a NumPy `ndarray`.

_Not_ added:

- `MaximisationProblem` support.
- Equality or inequality constraints.
- Any optimisations to prevent extra forward or first adjoint reruns (e.g. with second order methods which might call `TAOObjective.objective_gradient` followed by `TAOObjective.hessian`).


pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,361 @@
...
+                                                GradientNorm(), comm=comm)
+        M_inv_matrix.setOption(PETSc.Mat.Option.SYMMETRIC, True)
+        M_inv_matrix.setUp()
+        tao.setGradientNorm(M_inv_matrix)
+
+        if problem.bounds is not None:
+            def convert_bound(m, b, default):
+                if b is None:
+                    b = default
+                return m._ad_convert_type(b)
```
Ig-dolci, line 277: When using the method `m._ad_convert_type(b)`, do you expect to be returned a `Function` to have its data array equal to `b`?

jrmaddison: I believe here the bound value `b` should either have the same type as `m`, or be a scalar, yes. What is needed is "`m` dofs = `b`", but I don't see an easy way to do that via `OverloadedType` interfaces.

Ig-dolci: I think `_ad_convert_type` should have more options besides just `riesz_representation`. Perhaps something like `m._ad_convert_type(b, options={"convert_to_function": True})` could be useful. I will think more about this and discuss it with David. I will keep you updated.

dham:
I think that even if you are going to go via Numpy, this should be done by giving `OverloadedType` an `._ad_as_petsc_vec` method so that we can go back and do it properly for Firedrake Functions afterwards. As it stands, I think that the TAO interface would have to be substantially rewritten in order to support `OverloadedType`s that had a native mechanism for casting to PETSc Vec, and that seems unfortunate.

pyadjoint/optimization/tao_solver.py
jrmaddison: I've removed the use of `_ad_convert_type` here. If there is to be a check on bounds type then I think that should be in `OptimizationProblem.__check_arguments` -- at the moment this allows `int` or `float` bounds independent of the control type.

jrmaddison:
> I think that even if you are going to go via Numpy, this should be done by giving `OverloadedType` an `._ad_as_petsc_vec` method so that we can go back and do it properly for Firedrake Functions afterwards. As it stands, I think that the TAO interface would have to be substantially rewritten in order to support `OverloadedType`s that had a native mechanism for casting to PETSc Vec, and that seems unfortunate.

I think the 'one-or-more' pyadjoint interfaces mean that copying is going to be needed somewhere. e.g. there's nothing to stop a mixture of `AdjFloat`s and `firedrake.Function` controls. Maybe those could be PETSc copy operations if the backend supplies PETSc `Vec`s -- I suggested NumPy as an interface layer as that's close to what `OverloadedType`s already supply.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,357 @@
...
+
+    Args:
+        problem (MinimizationProblem): Defines the optimization problem to be
+            solved.
+        comm (petsc4py.PETSc.Comm or mpi4py.MPI.Comm): Communicator.
+        inner_product (str): Defines the Riesz map. Used to define the
+            inner product for derivatives with respect to the control.
+    """
+
+    def __init__(self, problem, parameters, *, comm=None, inner_product="L2"):
```
dham, line 197: `parameters` missing from Docstring.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,357 @@
...
+except ModuleNotFoundError:
+    PETSc = None
+
+__all__ = \
+    [
+        "TAOSolver"
+    ]
+
+
+class PETScVecInterface:
```
dham, line 22: This is missing docstrings which makes it really hard for me to understand.

dham:
OK, I can see what's going on here. I actually think it would be an easy fix to do this without the global gather, it just requires a few very simple methods to be added to Overloaded type, but I guess that can be delayed. Please add a health warning about the global gather in the class docstring, because the comment on the PR will not be visible to users.

pyadjoint/optimization/tao_solver.py
jrmaddison: Added

pyadjoint/optimization/tao_solver.py
jrmaddison: Documentation added. I've done my best to follow pyadjoint documentation conventions.

jrmaddison:
> OK, I can see what's going on here. I actually think it would be an easy fix to do this without the global gather, it just requires a few very simple methods to be added to Overloaded type, but I guess that can be delayed. Please add a health warning about the global gather in the class docstring, because the comment on the PR will not be visible to users.

It's not even a global gather, we just need to known the number of process local (owned) degrees of freedom, and `OverloadedType` doesn't provide this except indirectly via `_ad_convert_numpy`.

dham:
> > OK, I can see what's going on here. I actually think it would be an easy fix to do this without the global gather, it just requires a few very simple methods to be added to Overloaded type, but I guess that can be delayed. Please add a health warning about the global gather in the class docstring, because the comment on the PR will not be visible to users.
> 
> It's not even a global gather, we just need to known the number of process local (owned) degrees of freedom, and `OverloadedType` doesn't provide this except indirectly via `_ad_convert_numpy`.

OK, I'm confused again. I think the current implementation does a global gather onto a single numpy array. Is that not correct?

jrmaddison:
> > > OK, I can see what's going on here. I actually think it would be an easy fix to do this without the global gather, it just requires a few very simple methods to be added to Overloaded type, but I guess that can be delayed. Please add a health warning about the global gather in the class docstring, because the comment on the PR will not be visible to users.
> > 
> > 
> > It's not even a global gather, we just need to known the number of process local (owned) degrees of freedom, and `OverloadedType` doesn't provide this except indirectly via `_ad_convert_numpy`.
> 
> OK, I'm confused again. I think the current implementation does a global gather onto a single numpy array. Is that not correct?

You're correct. I'd missed (or forgotten) how `OverloadedType._ad_assign_numpy` and `OverloadedType._to_list` behave -- e.g. using a gather in `OverloadedType._to_list`.

This obviously won't scale, but I also don't see an alternative via the current `OverloadedType` interface.

jrmaddison:
Firedrake test required https://github.com/firedrakeproject/firedrake/pull/3657

pyadjoint/adjfloat.py
```diff
@@ -130,6 +130,12 @@ def _ad_str(self):
         """Return the string of the taped value of this variable."""
         return str(self.block_variable.saved_output)
 
+    def _ad_to_petsc(self):
+        NotImplementedError("It requires more thought to return a PETSc Vec from `AdjFloat`")
+
+    def _ad_from_petsc(self, vec):
+        NotImplementedError("It requires more thought.")
```
Ig-dolci, line 137:
```suggestion
        NotImplementedError("It requires more thought to return an `AdjFloat` from a PETSc Vec.")
```

pyadjoint/adjfloat.py
```diff
@@ -130,6 +130,12 @@ def _ad_str(self):
         """Return the string of the taped value of this variable."""
         return str(self.block_variable.saved_output)
 
+    def _ad_to_petsc(self):
+        NotImplementedError("It requires more thought to return a PETSc Vec from `AdjFloat`")
```
Ig-dolci, line 134:
```suggestion
        NotImplementedError("It requires more thought to return a PETSc Vec from AdjFloat.")
```

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,439 @@
...
+        """Communicator.
+        """
+
+        return self._comm
+
+    @property
+    def n(self):
+        """Total number of process local degrees of freedom, summed over all
+        variables.
+        """
```
Ig-dolci, line 82:
```suggestion
        """Total number of process local DoFs, summed over all variables."""
```

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,439 @@
...
+
+        return self._n
+
+    @property
+    def N(self):
+        """Total number of global degrees of freedom, summed over all
+        variables.
+        """
```
Ig-dolci, line 90:
```suggestion
        """Total number of global DoFs, summed over all variables."""
```

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,439 @@
...
+        for key, value in other.items():
+            self[key] = value
+
+
+class BoundType(Enum):
+    LOWER = 0
+    UPPER = 1
+
+
+class TAOObjective:
```
Ig-dolci, line 192: Need docstrings.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,439 @@
...
+    def __init__(self, rf):
+        self._reduced_functional = rf
+
+    @property
+    def reduced_functional(self):
+        return self._reduced_functional
+
+    def objective_gradient(self, M):
```
Ig-dolci, line 200: Docstring should specify whether the gradient is primal or dual.

Ig-dolci (changes requested):

pyadjoint/overloaded_type.py
```diff
@@ -142,6 +142,19 @@ def _ad_restore_at_checkpoint(self, checkpoint):
         """
         raise NotImplementedError
 
+    def _ad_assign(self, other):
+        """This method must be overridden.
+
+        The method should implement a routine for assigning other to `self`.
+
+        Note:
+            This method should not be employed for immutable types.
```
dham, line 151: I think this line here highlights an issue with this approach. `OverloadedType` actually conflates two sets of functionality:
1. Things needed in order to tape a variable.
2. Things needed in order to work with variables when computing taped operations (replay, tlm, adjoint etc.) and optimisation.

This functionality is part of 2 (it's not needed for taping). At that stage of operations, all variables are treated as immutable, so assignment looks like a strange operation at best. The fact that it doesn't work for some types is a further red flag.

Assignment itself is also not really what is meant here. The use case in question seems to be assigning a scalar to a vector, which is not well-defined (think about non-point-evaluation FunctionSpaces, for example). I think what is really meant here is interpolate. 

On the mutability point, I think this should be more like `_ad_init_object` and return a new value. I.e. the signature would be:
```
    @classmethod
    def _ad_interpolate(cls, obj, value):
```
and the required semantics would be to return a new object which is identical to `obj` except that its value approximates `value`.

dham (changes requested):

pyadjoint/overloaded_type.py
jrmaddison: `_ad_iadd` and `_ad_imul` are existing in-place operations. However I agree here an in-place operation is not needed. I think the easiest is just to change `VecInterface.to_petsc` to handle the scalar case using `VecSetValues`.

I also don't think scalar bounds are clearly defined. These could mean

- Bounds the dof values. I think this is in the intention, and is what `VecSetValues` would do.
- Bound function values. Too difficult, but see #145.
- Bound interpolated function values, but this is backend specific. Interpolation may not be meaningfully defined.

dham: This would bring us back to the option of requiring the user to specify bounds of the control type. I think this is probably the right thing -  Pyadjoint can't interpret the meaning of variables beyond basic Hilbert space operations (i.e. vector space operations + inner product). 

I guess the corner case is the no bounds case. We could either make that a user problem or we'd have to add `_ad` methods for producing variables that are element-wise minimal or maximal. That would be convenient for users but I'm not sure it smells good.

jrmaddison: > requiring the user to specify bounds of the control type

I'd definitely prefer that. Existing pyadjoint examples use scalars, though.

> I guess the corner case is the no bounds case.

That at least can be handled with `VecSetValues` and `np.finfo(PETSc.ScalarType).min`/`np.finfo(PETSc.ScalarType).max`.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,439 @@
...
+        return self._x
+
+    @staticmethod
+    def _prepared_bound(control, bound, bound_type):
+        if bound is None:
+            bound = {BoundType.LOWER: np.finfo(PETSc.ScalarType).min,
+                     BoundType.UPPER: np.finfo(PETSc.ScalarType).max}[bound_type]
+        bound_val = bound
+        bound = control.control._ad_copy()
+        bound._ad_assign(bound_val)
```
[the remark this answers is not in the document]
jrmaddison: Now applied using PETSc in `PETScVecInterface`.

pyadjoint/overloaded_type.py
jrmaddison: I've implemented the 'scalar bound = bound on all dofs' case using 'VecSet`'. This might still lead to some surprises depending on the basis, but I think is consistent with pyadjoint behaviour elsewhere.

pyadjoint/overloaded_type.py
```diff
@@ -240,6 +253,31 @@ def _ad_will_add_as_output(self):
         """
         return True
 
+    def _ad_to_petsc(self):
```
jrmaddison, line 243: @Ig-dolci I think this should have an optional `vec` argument, which will be set (and returned) if provided.

pyadjoint/overloaded_type.py
jrmaddison: Current pyadjoint behaviour is described here.

https://github.com/dolfin-adjoint/pyadjoint/blob/92121af97711be88ff6d7994c0c7469129b2fb25/pyadjoint/optimization/optimization_problem.py#L24-L28

jrmaddison: The actual check allows `None` and `int` as well.
https://github.com/dolfin-adjoint/pyadjoint/blob/92121af97711be88ff6d7994c0c7469129b2fb25/pyadjoint/optimization/optimization_problem.py#L51-L55

pyadjoint/overloaded_type.py
Ig-dolci: I see. In fact, `_ad_to_petsc` with `vec` optional argument will be more useful.

pyadjoint/adjfloat.py
```diff
@@ -130,6 +130,12 @@ def _ad_str(self):
         """Return the string of the taped value of this variable."""
         return str(self.block_variable.saved_output)
 
+    def _ad_to_petsc(self):
```
Ig-dolci, line 133:
```suggestion
    def _ad_to_petsc(self, vec=None):
```

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+
+        Y = Enlist(Y)
+        for iset, y in zip(self._isets, Y):
+            x_sub = x.getSubVector(iset)
+            if isinstance(y, Complex):
+                x_sub.set(y)
+            elif isinstance(y, OverloadedType):
+                y_vec = y._ad_to_petsc()
+                y_vec.copy(result=x_sub)
+                y_vec.destroy()
```
Ig-dolci, line 137:
```suggestion
                x_sub = y._ad_to_petsc(vec=x_sub)
```
Does the `_ad_to_petsc` with optional `vec` work better here?

pyadjoint/overloaded_type.py
```diff
@@ -240,6 +240,31 @@ def _ad_will_add_as_output(self):
...
```
Ig-dolci, line 243:
```suggestion
    def _ad_to_petsc(self, vec=None):
```

pyadjoint/overloaded_type.py
```diff
@@ -240,6 +240,31 @@ def _ad_will_add_as_output(self):
...
+        """This method must be overridden.
+
+        This method should implement a routine to return a new instance that is
+        a copy of the PETSc Vec associated with the overloaded object.
```
Ig-dolci, line 247:
```suggestion
        If the `vec` parameter is not provided, this method returns a new
        instance that is a copy of the PETSc Vec associated with the
        OverloadedType object. If the `vec` parameter is supplied, the PETSc
        Vec associated with the OverloadedType object is copied into the
        provided `vec`.

        Args:
            vec (PETSc.Vec, optional): A PETSc Vec to copy the data into.
```

pyadjoint/overloaded_type.py
Ig-dolci: I just rewrote `_ad_to_petsc` with `vec` as an optional argument in  [PR 3657](https://github.com/firedrakeproject/firedrake/pull/3657).

Ig-dolci (changes requested):

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+except ModuleNotFoundError:
+    PETSc = None
+
+__all__ = \
+    [
+        "TAOSolver"
+    ]
+
+
+def attach_destroy_finalizer(obj, *args):
```
dham, line 24: Comment please. This is black magic.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+    def finalize_callback(*args):
+        for arg in args:
+            if arg is not None:
+                arg.destroy()
```
dham, line 28: @JDBetteridge says this needs to be `.delayed_destroy`.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+        if comm is None:
+            comm = PETSc.COMM_WORLD
+        if hasattr(comm, "tompi4py"):
+            comm = comm.tompi4py()
+
+        vecs = tuple(x._ad_to_petsc() for x in X)
+        n = sum(vec.getLocalSize() for vec in vecs)
+        N = sum(vec.getSize() for vec in vecs)
+        _, isets = PETSc.Vec().concatenate(vecs)
+        for vec in vecs:
```
dham, line 61: If the destructor is really not necessary, let's not do it.

if it is necessary then `_ad_to_petsc()` should probably be a context manager so as to remove the footgun.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+            if isinstance(y, Complex):
+                x_sub.set(y)
+            elif isinstance(y, OverloadedType):
+                y._ad_to_petsc(vec=x_sub)
+            else:
+                raise TypeError(f"Unexpected type: {type(y)}")
+            x_sub.restoreSubVector(iset, x_sub)
+
+
+class PETScOptions:
```
dham, line 143: Pirate the PETSc Options handler from Firedrake.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+                value.
+        Returns:
+            AdjFloat: The value of the functional.
+            OverloadedType or Sequence[OverloadedType]: The (dual space)
+                derivative.
+        """
+
+        M = Enlist(M)
+        J = self.reduced_functional(tuple(m._ad_copy() for m in M))
+        _ = self.reduced_functional.derivative()
```
dham, line 223: Can you please raise an issue on Pyadjoint pointing out that this hack is here.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+
+    @property
+    def reduced_functional(self):
+        """:class:`.ReducedFunctional`. Defines the forward, and used to
+        compute derivative information.
+        """
+
+        return self._reduced_functional
+
+    def objective_gradient(self, M):
```
dham, line 209: `M` should probably be lower case.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+        M = Enlist(M)
+        M_dot = Enlist(M_dot)
+        _ = self.reduced_functional(tuple(m._ad_copy() for m in M))
+        _ = self.reduced_functional.derivative()
+        _ = self.reduced_functional.hessian(tuple(m_dot._ad_copy() for m_dot in M_dot))
+        ddJ = tuple(m.control.block_variable.hessian_value
+                    for m in self.reduced_functional.controls)
+        return M.delist(ddJ)
+
+    def new_M(self):
```
dham, line 251: `new_control_variable`

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+        Returns:
+            OverloadedType or Sequence[OverloadedType]: New variable or
+                variables suitable for storing a control value.
+        """
+
+        # Not initialized to zero
+        return tuple(m.control._ad_copy()
+                     for m in self.reduced_functional.controls)
+
+    def new_M_dual(self):
```
dham, line 264: `new_dual_control_variable`.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+        for m in self.reduced_functional.controls:
+            bv = m.control.block_variable
+            if bv.adj_value is not None:
+                M_dual.append(bv.adj_value)
+            elif bv.hessian_value is not None:
+                M_dual.append(bv.hessian_value)
+            else:
+                raise RuntimeError("Unable to instantiate new dual space "
+                                   "object")
+        return tuple(m_dual._ad_copy() for m_dual in M_dual)
```
dham, line 287: Please raise this hack as an issue too.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+            J_val, dJ = taoobjective.objective_gradient(M)
+            to_petsc(g, dJ)
+            return J_val
+
+        tao.setObjectiveGradient(objective_gradient, None)
+
+        def hessian(tao, x, H, P):
+            H.getPythonContext().set_M(x)
+
+        class Hessian:
```
dham, line 339: Docstring saying what interface this implements.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+                if self._shift != 0.0:
+                    y.axpy(self._shift, x)
+
+        H_matrix = PETSc.Mat().createPython(((n, N), (n, N)),
+                                            Hessian(), comm=comm)
+        H_matrix.setOption(PETSc.Mat.Option.SYMMETRIC, True)
+        H_matrix.setUp()
+        tao.setHessian(hessian, H_matrix)
+
+        class GradientNorm:
```
dham, line 368: also docstring saying what this thing implements.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+                lbs.append(lb)
+                ubs.append(ub)
+
+            lb_vec = vec_interface.new_petsc()
+            ub_vec = vec_interface.new_petsc()
+            to_petsc(lb_vec, lbs)
+            to_petsc(ub_vec, ubs)
+            tao.setVariableBounds(lb_vec, ub_vec)
+            lb_vec.destroy()
+            ub_vec.destroy()
```
dham, line 401: please don't destroy unless needed.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+        options.update(parameters)
+        tao.setOptionsPrefix(options.options_prefix)
+
+        tao.setFromOptions()
+
+        if tao.getType() in {PETSc.TAO.Type.LMVM, PETSc.TAO.Type.BLMVM}:
+            class InitialHessian:
+                pass
+
+            class InitialHessianPreconditioner:
```
dham, line 413: explain that this is a PC.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+                def apply(self, pc, x, y):
+                    X = taoobjective.new_M_dual()
+                    from_petsc(x, X)
+                    assert len(taoobjective.reduced_functional.controls) == len(X)
+                    X = tuple(m._ad_convert_type(x, options=convert_options)
+                              for m, x in zip(taoobjective.reduced_functional.controls, X))
+                    to_petsc(y, X)
+
+            B_0_matrix = PETSc.Mat().createPython(((n, N), (n, N)),
```
dham, line 422: Have mercy on your reader and explain that B_0 is H_0

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+        self._taoobjective = taoobjective
+        self._vec_interface = vec_interface
+        self._tao = tao
+        self._x = x
+
+        attach_destroy_finalizer(
+            self, tao, H_matrix, M_inv_matrix, B_0_matrix_pc, B_0_matrix, x)
+
+    @property
+    def taoobjective(self):
```
dham, line 456: PEP8 name should have underscore.

pyadjoint/optimization/tao_solver.py
```diff
@@ -0,0 +1,491 @@
...
+            OverloadedType or tuple[OverloadedType]: The solution.
+        """
+        M = tuple(
+            m.tape_value()._ad_copy()
+            for m in self.taoobjective.reduced_functional.controls)
+        self._vec_interface.to_petsc(self.x, M)
+        self.tao.solve()
+        self._vec_interface.from_petsc(self.x, M)
+        if self.tao.getConvergedReason() <= 0:
+            raise RuntimeError("Convergence failure")
```
dham, line 490: Is this the conventional exception for solver fails?

dham (changes requested):

pyadjoint/overloaded_type.py
jrmaddison: As discussed in the meeting, this should be fixed separately from this PR.

pyadjoint/optimization/tao_solver.py
jrmaddison: Done.

pyadjoint/optimization/tao_solver.py
jrmaddison: Not available, but the possible deadlock is now documented.

pyadjoint/optimization/tao_solver.py
jrmaddison: Removed.

pyadjoint/optimization/tao_solver.py
jrmaddison: Done.

pyadjoint/optimization/tao_solver.py
jrmaddison: #164

pyadjoint/optimization/tao_solver.py
jrmaddison: Changed to `m`. Also changed `X` to `x` and `Y` to `y`.

pyadjoint/optimization/tao_solver.py
jrmaddison: Done.

pyadjoint/optimization/tao_solver.py
jrmaddison: Done.

pyadjoint/optimization/tao_solver.py
jrmaddison: #165

pyadjoint/optimization/tao_solver.py
jrmaddison: Done.

pyadjoint/optimization/tao_solver.py
jrmaddison: Done.

pyadjoint/optimization/tao_solver.py
jrmaddison: Done.

pyadjoint/optimization/tao_solver.py
jrmaddison: Done.

pyadjoint/optimization/tao_solver.py
jrmaddison: Added reference to Nocedal and Weight for the notation.

pyadjoint/optimization/tao_solver.py
jrmaddison: Fixed, although I followed `ROLSolver.rolobjective`.

pyadjoint/optimization/tao_solver.py
jrmaddison: Updated to use a `TAOConvergenceError` with a message similar to Firedrake linear solve failures.

pyadjoint doesn't seem to raise exceptions on for optimizer failures in the SciPy or ROL interfaces. I'm pretty sure that's unsafe for the SciPy interface, #166.

pyadjoint/optimization/tao_solver.py
jrmaddison: This is intended to work around the following, which leaks memory on my machine (Ubuntu 24.04, run on two processes)
```
from firedrake import *
from firedrake.petsc import garbage_cleanup

import numpy as np

from itertools import count
import gc
import weakref


class Mat:
    def __init__(self, space):
        self._x = Function(space)
        self._test = TestFunction(space)
        self._memory = np.full(1000000, 1, dtype=np.uint8)

    def mult(self, x, y):
        with self._x.dat.vec_ro:
            x.copy(result=x_v)
        y_c = assemble(inner(self._x, self._test) * dx)
        with y_c.dat.vec_ro as y_v:
            y_v.copy(result=y)


def attach_destroy_finalizer(obj, *args):
    def finalize_callback(*args):
        for arg in args:
            if arg is not None:
                arg.destroy()

    finalize = weakref.finalize(obj, finalize_callback,
                                *args)
    finalize.atexit = False


mesh = UnitSquareMesh(10, 10)
space = FunctionSpace(mesh, "Lagrange", 1)
u = Function(space, name="u")
with u.dat.vec_ro as u_v:
    n, N = u_v.getSizes()


class Test:
    pass


for _ in count():
    t = Test()
    A = Mat(space)
    mat = PETSc.Mat().createPython(((n, N), (n, N)), A, comm=u.comm)
    # attach_destroy_finalizer(t, mat)  # Uncomment to fix leak

    del t
    gc.collect()
    garbage_cleanup(u.comm)
```
Arguably it would be better to track down the source of the leak -- but the case where I've hit this was much harder to debug than the above, hence the defensive use of `destroy`. The main downside is if properties are accessed and retained, e.g.
```
tao_solver = TAOSolver(...)
tao = tao_solver.tao
del tao_solver
# TaoDestroy has been called for tao
```

jrmaddison: The previous example leaks because `garbage_cleanup(u.comm)` doesn't call `PETSc.garbage_cleanup(u.comm)` (is this a separate bug?). However a bit more digging leads to this example, which leaks for me with two processes.
```
from firedrake import *
from firedrake.petsc import garbage_cleanup

import petsc4py.PETSc as PETSc

import gc
from itertools import count


class Mat:
    def mult(self, A, x, y):
        x.copy(result=y)


mesh = UnitSquareMesh(10, 10)
space = FunctionSpace(mesh, "Lagrange", 1)
u = Function(space, name="u")
with u.dat.vec_ro as u_v:
    n, N = u_v.getSizes()
b = Function(space).interpolate(Constant(1))


for i in count():
    A = Mat()
    mat = PETSc.Mat().createPython(((n, N), (n, N)), A, comm=u.comm)
    ksp = PETSc.KSP().createPython()
    ksp.setOperators(mat)
    ksp.setTolerances(rtol=0.1)
    ksp.setType(PETSc.KSP.Type.CG)
    with b.dat.vec as b_v, Function(space).dat.vec as u_v:
        ksp.solve(b_v, u_v)

    # Uncomment to fix leak
    # ksp.destroy()
    # mat.destroy()

    del mat, ksp

    # Increase to speed up leak
    j = 1
    if i % j == 0:
        print(f"{i=}", flush=True)
        gc.collect()
        garbage_cleanup(u.comm)
        PETSc.garbage_cleanup(u.comm)
```

dham (approved):

PR #147, opened 2024-05-13
Checkpoint clean based on the OverloadedType Function response

Ig-dolci:


pyadjoint/overloaded_type.py
```diff
@@ -329,6 +334,9 @@ def create_block_variable(self):
         block_variable.floating_type = True
         return block_variable
 
+    def _ad_checkpoint_to_clear(self, to_keep=None):
```
Ig-dolci, line 337:
```suggestion
    def _ad_is_to_clear_checkpoint(self, to_keep=None):
```

PR #156, opened 2024-06-27
abstract reduced functional

dham:
Refactor the reduced functional as a subclass of an abstract class with a more consistent interface.


jorgensd (approved):
LGTM

JHopeCollins:
> Does this resolve #164 and #165?

Yes it will.

PR #157, opened 2024-06-28
Get OverloadType PETSc.vec

Ig-dolci:
This PR adds an `OverloadType` method to get the `PETSc.petsc4py.Vec` object.

### DIFF
--- a/pyadjoint/overloaded_type.py
+++ b/pyadjoint/overloaded_type.py
@@ -323,7 +323,7 @@ class OverloadedType(object):
     def _ad_petsc_vec_read_only(self):
         """This method must be overridden.
 
-        The method should implement a routine to return a read only PETSc Vec object.
+        The method should implement a routine to return a read only a PETSc Vec object.
 
         Returns:
             PETSc.Vec: A PETSc Vec object containing the data of the overloaded.
@@ -334,7 +334,7 @@ class OverloadedType(object):
     def _ad_petsc_vec_write_only(self):
         """This method must be overridden.
 
-        The method should implement a routine to return a write only PETSc Vec object.
+        The method should implement a routine to write only a PETSc Vec object.
 
         Returns:
             PETSc.Vec: A PETSc Vec object containing the data of the overloaded.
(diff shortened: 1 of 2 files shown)


pyadjoint/overloaded_type.py
```diff
@@ -320,6 +320,28 @@ def _ad_str(self):
         """
         return str(self)
 
+    def _ad_petsc_vec_read_only(self):
+        """This method must be overridden.
```
dham, line 324: I think that in this case this overstates the case.
```suggestion
        """This method should be overwritten by types which can return a PETSc Vec.
```

pyadjoint/overloaded_type.py
```diff
@@ -320,6 +320,28 @@ def _ad_str(self):
...
+        The method should implement a routine to return a read only a PETSc Vec object.
+
+        Returns:
+            PETSc.Vec: A PETSc Vec object containing the data of the overloaded.
+
+        """
+        raise NotImplementedError
+
+    def _ad_petsc_vec_write_only(self):
+        """This method must be overridden.
```
dham, line 335: as above.

pyadjoint/overloaded_type.py
```diff
@@ -320,6 +320,28 @@ def _ad_str(self):
...
```
Ig-dolci, line 335:
```suggestion
        """This method should be overwritten by types which can return a PETSc Vec.
```

pyadjoint/overloaded_type.py
Ig-dolci: done.

jrmaddison:
Should document that these are context managers, and also document what is expected on entry/exit (ghost updates? other data consistency updates?).

Ig-dolci:
> Should document that these are context managers, and also document what is expected on entry/exit (ghost updates? other data consistency updates?).

Yes. It is a good point. I will study more about the management of Petsc vecs and try to improve the docs.

pyadjoint/overloaded_type.py
```diff
@@ -320,6 +320,42 @@ def _ad_str(self):
         """
         return str(self)
 
+    def _ad_petsc_vec_read_only(self):
```
Ig-dolci, line 323: Should I consider adding an optional argument like with_halos? Or does it not make sense?

Ig-dolci:
This PR is not necessary anymore. OverloadedTypes to copy and update PETSc Vecs are available at the [PR 143](https://github.com/dolfin-adjoint/pyadjoint/pull/143).

Issue #170, opened 2024-10-01
`ReducedFunctional.optimize_tape` should replace `self.tape` with the optimized tape without modifying the original `self.tape`.


JHopeCollins:
This means that if you call `ReducedFunctional.optimize_tape` on one `ReducedFunctional`, it will not modify the tape of any other `ReducedFunctional`.

Potentially the API should change to the `optimize_tape` returns the new (optimized) tape, rather than modifying `self.tape`.

PR #171, opened 2024-10-11
`ReducedFunctional.optimize_tape` will optionally _replace_ self.tape with the optimized tape, without modifying original tape.

JHopeCollins:
This PR addresses #170 

Currently `ReducedFunctional.optimize_tape` will modify `self.tape` in-place. This means that if anything else is using the same tape, then they now also have a modified tape.

This PR adds an optional `replace` argument to `optimize_tape` to give the option of copying the tape and replacing `self.tape` with the optimized tape, without modifying the original tape. The optimized tape is also returned (whether a copy was made or not).
I did it this way so that the default call has the same behaviour as before, so it shouldn't break any existing code - unless someone is relying on it returning `None`, but seeing as this is implicit and non-documented behaviour I don't think we care.

@dham also suggested that the default behaviour should change so that it doesn't modify `self.tape`, but just returns the optimized tape. Is this something we want to do? If so, should it raise a warning for now that the default behaviour will change in the future, and delay making the change for a bit?
I think if this behaviour does change, it would be good to have an `optimize_tape(modify=False)` argument to optionally still update `self.tape`, otherwise we'll probably see `Jhat.tape = Jhat.optimize_tape()` popping up all over.

### DIFF
--- a/pyadjoint/control.py
+++ b/pyadjoint/control.py
@@ -80,6 +80,7 @@ class Control(object):
     def fetch_numpy(self, value):
         return self.control._ad_to_list(value)
 
+    # TODO: This should be self.block_variable.checkpoint?
     def copy_data(self):
         return self.control._ad_copy()
 
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -231,6 +231,7 @@ class ReducedFunctional(object):
             controls=self.controls,
             functionals=[self.functional]
         )
+        return self.tape
 
     def marked_controls(self):
         return marked_controls(self)
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -77,7 +77,6 @@ class ReducedFunctional(object):
         if not isinstance(functional, OverloadedType):
             raise TypeError("Functional must be an OverloadedType.")
         self.functional = functional
-        self.tape = get_working_tape() if tape is None else tape
         self.controls = Enlist(controls)
         self.derivative_components = derivative_components
         self.scale = scale
@@ -87,6 +86,13 @@ class ReducedFunctional(object):
         self.derivative_cb_post = derivative_cb_post
         self.hessian_cb_pre = hessian_cb_pre
         self.hessian_cb_post = hessian_cb_post
+
+        tape = get_working_tape() if tape is None else tape
+        self.tape = tape.copy()
+        self.tape.optimize(
+            controls=self.controls,
+            functionals=[self.functional]
+        )
 
         if self.derivative_components:
             # pre callback
@@ -224,13 +230,9 @@ class ReducedFunctional(object):
 
         return func_value
 
-    def optimize_tape(self, replace=False):
-        if replace:
-            self.tape = self.tape.copy()
-        self.tape.optimize(
-            controls=self.controls,
-            functionals=[self.functional]
-        )
+    def optimize_tape(self):
+        # Tape already optimized in __init__
+        # TODO: What should we do here now?
         return self.tape
 
     def marked_controls(self):
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -224,7 +224,9 @@ class ReducedFunctional(object):
 
         return func_value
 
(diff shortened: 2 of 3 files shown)


jrmaddison:
> I think if this behaviour does change, it would be good to have an optimize_tape(modify=False) argument to optionally still update self.tape, otherwise we'll probably see Jhat.tape = Jhat.optimize_tape() popping up all over.

Is optimizing the tape actually expensive (it involves just a few passes over the tape)? I'd expect it's cheap enough (compared e.g. to a gradient calculation) that is isn't necessary to cache the optimized tape.

JHopeCollins:
> > I think if this behaviour does change, it would be good to have an optimize_tape(modify=False) argument to optionally still update self.tape, otherwise we'll probably see Jhat.tape = Jhat.optimize_tape() popping up all over.
> 
> Is optimizing the tape actually expensive (it involves just a few passes over the tape)? I'd expect it's cheap enough (compared e.g. to a gradient calculation) that is isn't necessary to cache the optimized tape.

The optional argument to update `self.tape` is so that you can still have (close to) the same behaviour as you have currently, i.e. `optimize_tape` actually does modify `self.tape`.

Otherwise anywhere that `Jhat.optimize_tape()` is currently called will inevitably get replaced with `Jhat.tape = Jhat.optimize_tape()`. This feels like a bit of a code smell because what if at some point the implementation changes and there's other things that `ReducedFunctional` needs to do if `self.tape` changes?

I guess the answer in that case would be to make `ReducedFunctional.tape` a `property`, but the `Jhat.tape = Jhat.optimize_tape()` still feels clunky compared to just having `Jhat.optimize_tape(modify=True)` (or whatever you want to call the argument).

jrmaddison:
What I mean is that, since it's (probably) cheap, the activity analysis can be performed each time a derivative is computed. i.e. I think it makes more sense as an argument, `ReducedFunctional.derivative(..., optimize=True)`, than as a method modifying the `ReducedFunctional`. The optimized version of the tape could easily be cached if that's still wanted, but not as `ReducedFunctional.tape`.

Modifying `ReducingFunctional.tape` probably also breaks Hessian actions.

dham:
> Modifying `ReducingFunctional.tape` probably also breaks Hessian actions.

Can you explain this? I don't understand why it would.

jrmaddison:
> Can you explain this? I don't understand why it would.

You're correct, I've worked through the second order and I think it's OK.

JHopeCollins:
I've tested out having `ReducedFunctional.__init__` take a copy of the tape and optimize it, so that `ReducedFunctional` only ever has a view over the most minimal part of the tape required.

The main issue I ran into is that `Tape.copy` creates an incomplete copy that only works in the most simple situations, mainly down to not copying checkpoints.  The current `copy` doesn't transfer over the checkpoint manager, so there were errors anywhere a manager was used. I made two changes in `Tape.copy` to "fix" this. Better fixes are almost certainly needed.

1. The checkpoint manager is attached to the new tape. This fixed the errors for in-memory checkpointing.
The private `_checkpoint_manager` member is set explicitly because `Tape.enable_checkpointing` will throw an error if the tape already has blocks on. The checkpoint manager now has a reference to the wrong tape (but at the time of the copy that wrong tape still has all the right blocks for the new `Tape` in it). I imagine this is a Very Bad Thing to do, open to suggestions for the Right Thing.
I could have a go at implementing a `copy` for `CheckpointManager`, I imagine it would need to take a `tape` argument and try to copy only the bits of the schedule that correspond to that tape (and error if it doesn't have something it needs).

2. Any items in the `package_data` that don't have a `copy` method just get transferred to the `package_data` for the new `Tape`. This fixes the errors for disk checkpointing.
This was to get around the fact that `firedrake.DiskCheckpointer` doesn't implement `copy`. Again, open to suggestions for the Right Thing.

JHopeCollins:
Closing this PR as it isn't clear how to deal with the issues in the comment above, particularly what it means to copy and optimize a tape with a checkpoint manager.
The aim of this PR (multiple `ReducedFunctionals` with very localised and independent tapes) can be achieved with liberal use of `set_working_tape` and `stop_annotating`.

PR #176, opened 2024-11-14
Enable pytest the tests

Ig-dolci:
Enable local `pytest tests/`. Currently, only `python -m pytest` works.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,9 +19,6 @@ classifiers = [
   "Programming Language :: Python",
 ]
 
-[tool.pytest.ini_options]
-pythonpath = ["."]
-
 [project.urls]
 Repository = "https://github.com/dolfin-adjoint/pyadjoint.git"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,6 @@ classifiers = [
 ]
 
 [tool.pytest.ini_options]
-# This makes the root directory available in PYTHONPATH for pytest runs.
 pythonpath = ["."]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,10 @@ classifiers = [
   "Programming Language :: Python",
 ]
 
+[tool.pytest.ini_options]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -19,6 +19,10 @@ classifiers = [
   "Programming Language :: Python",
 ]
 
+[tool.pytest.ini_options]
+# This makes the root directory available in PYTHONPATH for pytest runs.
```
Ig-dolci, line 23:
```suggestion
```

Ig-dolci:
> Would specifying `testpaths` (https://docs.pytest.org/en/stable/reference/customize.html) work instead?
> 
> This looks quite magic

Hum... https://docs.pytest.org/en/stable/reference/customize.html does not work here.

pyproject.toml
```diff
@@ -19,6 +19,9 @@ classifiers = [
   "Programming Language :: Python",
 ]
 
+[tool.pytest.ini_options]
+pythonpath = ["."]
```
connorjward, line 23: So specifically I was suggesting doing this:
```suggestion
testpaths = ["tests"]
```
Does this not do the right thing?

Ig-dolci: According to https://docs.pytest.org/en/stable/reference/customize.html, this should work but does not. At least here.

connorjward: I think I have found a less magic solution for this. From the [pytest docs](https://docs.pytest.org/en/stable/explanation/pythonpath.html) it says
> It is highly recommended to arrange your test modules as packages by adding __init__.py files to your directories containing tests. This will make the tests part of a proper Python package, allowing pytest to resolve their full name (for example tests.core.test_core for test_core.py inside the tests.core package).

Adding a `tests/__init__.py` file made things work for me. Can you try the same? If we decide that that is the solution can you include a comment in the `__init__.py` file linking to the pytest docs above?

Ig-dolci: Yes. It is working. Thanks.

connorjward (approved):

PR #177, opened 2024-11-19
Allow using `set_working_tape` as a function decorator

JHopeCollins:
## Description
This PR uses the `ContextDecorator` class allow `set_working_tape` and `stop_annotating` to be used as function decorators.

The `no_annotations` decorator is now implemented as an instance of `stop_annotating`. Because `no_annotations` contexts could be nested, we need to keep track of the original annotation states at entry to each context separately. This is done by pushing/popping from a stack in `__enter__` and `__exit__`.

### `set_working_tape` decorator
The two functions below are equivalent:
```python
@set_working_tape()
def decorated_function(*args, **kwargs):
    # do something here
    return ReducedFunctional(functional, control)

def context_function(*args, **kwargs):
    with set_working_tape():
        # do something here
        return ReducedFunctional(functional, control)
```

### `stop_annotating` decorator
The two functions below are equivalent:
```python
@stop_annotating()
def decorated_function(*args, **kwargs):
    # do something here

@no_annotations
def decorated_function(*args, **kwargs):
    # do something here
```


pyadjoint/tape.py
```diff
@@ -28,9 +27,11 @@ def continue_annotation():
...
 
 
-class set_working_tape(object):
-    """A context manager whithin which a new tape is set as the working tape.
-       This context manager can also be used in an imperative manner.
+class set_working_tape(ContextDecorator):
+    """Set a new tape as the working tape. This can be used either:
+       1) as a free function to replace the working tape,
+       2) as a context manager within which a new tape is set as the working tape
+       3) as a function decorator so that the new tape is set only inside the function.
```
connorjward, line 34:
```suggestion
    """Set a new tape as the working tape.
    
    This class can be used in three ways:
       1) as a free function to replace the working tape,
       2) as a context manager within which a new tape is set as the working tape,
       3) as a function decorator so that the new tape is set only inside the function.
```

Docstring best practice is to have a short description followed by a blank line.

pyadjoint/tape.py
```diff
@@ -101,15 +125,7 @@ def __exit__(self, *args):
...
-def no_annotations(function):
-    """Decorator to turn off annotation for the decorated function."""
-
-    @wraps(function)
-    def wrapper(*args, **kwargs):
-        with stop_annotating():
-            return function(*args, **kwargs)
-
-    return wrapper
+no_annotations = stop_annotating()
```
connorjward, line 128:
```suggestion
no_annotations = stop_annotating()
"""Decorator to turn off annotation for the decorated function."""
```

pyadjoint/tape.py
```diff
@@ -82,17 +100,23 @@ class stop_annotating(object):
     modified variables at the end of the context manager. """
 
     def __init__(self, modifies=None):
-        global _annotation_enabled
         self.modifies = modifies
-        self._orig_annotation_enabled = _annotation_enabled
+        # the `no_annotations` context manager could be nested,
+        # so we need a stack to keep track of the original states.
```
connorjward, line 105: Is this really necessary? I helped implement this originally and I thought that using `_orig_annotation_enabled` was enough to simulate a stack.

connorjward (changes requested):

pyadjoint/tape.py
JHopeCollins: I did not know you could do that, that's very useful.

pyadjoint/tape.py
JHopeCollins: Previously, the wrapper that `no_annotations` returned would instantiate a new instance of `stop_annotating` for every call of the decorated function. The "stack" was just lots of different instances of `stop_annotating`.

Now, because `no_annotations` is a single instance of `stop_annotating`, if you nest two functions decorated with `no_annotation`, then they both enter/exit the _same_ instance of `stop_annotating`. Hence we need an explicit stack.

connorjward: OK that makes some sense.

PR #185, opened 2024-12-09
Unit test for use of a Riesz map with TAO

jrmaddison:
Tests the use of an $L^2$ Riesz map by a pyadjoint `TAOSolver`, versus the use of an an $l^2$ Riesz map with an $L^2$ orthonormal basis for the control space. The latter is implemented using a mass matrix inverse square root action performed using a `SLEPc` `MFN`.

Currently `TAOSolver` only handles Riesz maps for the `LMVM` and `BLMVM` `tao_type`s. The unit test `test_simple_inversion_riesz_representation` can be expanded if/when Riesz map functionality with other `tao_type`s is added.

### DIFF
--- a/tests/firedrake_adjoint/test_optimisation.py
+++ b/tests/firedrake_adjoint/test_optimisation.py
@@ -210,6 +210,8 @@ def transform(v, transform_type, *args, mfn_parameters=None, **kwargs):
 
         options.set_from_options(mfn)
         mfn.setUp()
+        if mfn.getFN().getType() != SLEPc.FN.Type.SQRT:
+            raise ValueError("Invalid FN type")
 
         with v.dat.vec_ro as v_v:
             x = v_v.copy()
--- a/tests/firedrake_adjoint/test_optimisation.py
+++ b/tests/firedrake_adjoint/test_optimisation.py
@@ -325,7 +325,7 @@ def test_simple_inversion_riesz_representation(tao_type):
                             mfn_parameters=mfn_parameters)
     assert_allclose(x_transform.dat.data, source_ref.dat.data, rtol=1e-2)
 
-    assert solver_transform.tao.getIterationNumber() <= solver.tao.getIterationNumber()
+    assert solver.tao.getIterationNumber() <= solver_transform.tao.getIterationNumber()
 
 
 def test_tao_bounds():
--- a/tests/firedrake_adjoint/test_optimisation.py
+++ b/tests/firedrake_adjoint/test_optimisation.py
@@ -1,12 +1,21 @@
 import pytest
 pytest.importorskip("firedrake")
 
+from enum import Enum, auto
 from numpy.testing import assert_allclose
 import numpy as np
+try:
+    import petsc4py.PETSc as PETSc
+except ModuleNotFoundError:
+    PETSc = None
+try:
+    import slepc4py.SLEPc as SLEPc
+except ModuleNotFoundError:
+    SLEPc = None
 from firedrake import *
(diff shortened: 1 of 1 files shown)


jrmaddison:
Note that this isn't a great implementation of the transform -- presumably it's better to compute the action of the square root of the mass matrix (rather than the mass matrix inverse), and you might want to precondition (e.g https://tlm-adjoint.github.io/autoapi/tlm_adjoint/firedrake/block_system/index.html#tlm_adjoint.firedrake.block_system.WhiteNoiseSampler) or use a Cholesky factorization -- but

tests/firedrake_adjoint/test_optimisation.py
```diff
@@ -157,13 +164,177 @@ def test_tao_simple_inversion(minimize, riesz_representation):
...
+
+    solver_transform = TAOSolver(
+        MinimizationProblem(rf_transform), tao_parameters,
+        convert_options={"riesz_representation": "l2"})
+    x_transform = transform(solver_transform.solve(), TransformType.PRIMAL,
+                            riesz_representation,
+                            mfn_parameters=mfn_parameters)
+    assert_allclose(x_transform.dat.data, source_ref.dat.data, rtol=1e-2)
+
+    assert solver_transform.tao.getIterationNumber() <= solver.tao.getIterationNumber()
```
Ig-dolci, line 328: Let me see if I understood right. This test considers the case where the optimisation solver uses the derivatives instead of the gradients. Then, `transform` (using mass Matrix) is applied to convert the `TaoSolver` results to the primal space. Is that right? If yes, That is extendable to other `tao_type`s, right?

jrmaddison: Short answer: `transform` applies a mass matrix inverse square root, so that the control is expanded in an $L^2$ orthonormal basis, and so that the $l^2$ inner product is the same as the $L^2$ inner product. This means the derivative is the same as the gradient, and we get mesh-independent optimizers using $l^2$ inner products. There's a minor hack where the dofs in the new basis are stored in `Function`s/`Cofunction`s.

Long answer: Assuming $L^2$ is the relevant inner product, and considering real spaces.

FEM encounters issues with Riesz maps because we choose a basis with local support, but sacrifice $L^2$ orthogonality. There's nothing to stop us choosing a basis which is $L^2$ orthogonal, it's just that the basis no longer has local support (in general).

However we can still transform from an expansion in the usual FEM basis to an expansion in an $L^2$ orthonormal basis,

$$\hat{\phi} = J \tilde{\phi},$$

where overhat means dofs in some $L^2$ orthonormal basis and tilde means dofs in the original FEM basis. We want

$$\hat{\phi}_2^T \hat{\phi}_1 = \tilde{\phi}_2^T J^T J \tilde{\phi}_1 = \tilde{\phi}_2^T M \tilde{\phi}_1,$$

for any two $\tilde{\phi}_1$, $\tilde{\phi}_2$, where $M$ is the mass matrix. This means $J$ needs to satisfy

$$J^T J = M.$$

So we can e.g. choose $J$ to be the principal square root of the mass matrix, or the upper Cholesky factor, or make other choices -- each corresponding to a possibly different choice of $L^2$ orthonormal basis.

Fortunately, as usual, we don't need $J$, only it's action, which we can compute matrix free. The test computes the matrix-free square root action using  SLEPc.

Ig-dolci: I got it. This looks right to me.

Ig-dolci (approved):

PR #187, opened 2024-12-13
Optional garbage collection and CheckpointManager._global_deps

Ig-dolci:
### **PR Description**  

- **Enable the user to apply the garbage collection if necessary:** 
This PR introduces garbage collection optional support during checkpointing to enable the user to handle the lack of Python to properly track and clean up checkpoint objects in memory. 

_Experiment details used to test the garbage collection imposed manually_

- **Degree of Freedom (DoFs):** 40,401  
- **Test Type:** Burgers test  
- **Total Steps:** 1,000  

The black curve represents the scenario with garbage collection enabled, while the blue curve shows the case without garbage collection during checkpointing.
Memory Usage

- **Checkpoint Manager _global_deps:**  
  A private attribute, `_global_deps`, is introduced in the `CheckpointManager` class. This attribute stores dependencies that are used at each time step and are not time-dependent.  
  - If a block variable is included in `_global_deps`, it will not be cleaned during checkpointing. This prevents unnecessary cleanup and re-creation of checkpoints for dependencies that do not change with time.


pyadjoint/tape.py
```diff
@@ -312,12 +312,19 @@ def enable_checkpointing(self, schedule):
         Args:
             schedule (checkpoint_schedules.schedule): A schedule provided by the
             checkpoint_schedules package.
+            gc_collect_opts (dict): A dictionary of options to be passed to the
+            garbage collector. The keys considered here is `timestep_frequency`.
+            Optionally the user can provide the `generation` key. The default value
```
Ig-dolci, line 317: Add generation and say correctly about the key for checkpointing and the generation key.

Ig-dolci: Add two args (gc_timestep_frequency and gc_generation) instead of adding a dictionary.

pyadjoint/checkpointing.py
```diff
@@ -164,11 +172,22 @@ def _(self, cp_action, timestep):
             ):
                 for package in self.tape._package_data.values():
                     package.continue_checkpointing()
+            if timestep == 1:
+                # Store the supposed global dependencies.
```
Ig-dolci, line 176:
```suggestion
                # Store the possible global dependencies.
```

jrmaddison:
Is it equivalent to instead drop zero output `Block`s on the tape for garbage collection? Or is this slightly different?

Ig-dolci:
> Is it equivalent to instead drop zero output `Block`s on the tape for garbage collection? Or is this slightly different?

I believe it is different. I noticed that mainly during the recomputation process, memory usage kept growing, even after I cleared the checkpoint using block_variable._checkpoint = None. After some discussions here, my hypothesis is that Python might not be tracking all objects in memory properly. So, I am only allowing the user to employ the garbage collector manually, which looks like is helping.

pyadjoint/checkpointing.py
```diff
@@ -379,6 +416,13 @@ def _(self, cp_action, progress_bar, markings, functional=None, **kwargs):
                         out.reset_variables(("adjoint", "hessian"))
                     if cp_action.clear_adj_deps and out not in to_keep:
                         out._checkpoint = None
+            if self.self._gc_time_frequency:
```
Ig-dolci, line 419:
```suggestion
            if self._gc_time_frequency:
```

pyadjoint/tape.py
```diff
@@ -787,19 +794,28 @@ def copy(self, blocks=None):
...
         return out
 
-    def checkpoint(self, checkpointable_state, adj_dependencies):
+    def checkpoint(self, checkpointable_state, adj_dependencies, global_deps):
         """Store a copy of the checkpoints in the checkpointable state.
 
+        If a checkpoint is not time-dependent, i.e., it is in global dependency,
+        we only point to the checkpoint in the checkpointable state, avoiding
+        unnecessary copying.
+
```
Ig-dolci, line 803:
```suggestion
```

pyadjoint/tape.py
```diff
@@ -787,19 +794,28 @@ def copy(self, blocks=None):
...
             to compute the adjoint of a timestep.
+            global_deps (set): The set of global dependencies. These dependencies should
+            not be time-dependent and should be in the checkpointable states at every time step.
         """
         with stop_annotating():
             if checkpointable_state:
                 for var in self.checkpointable_state:
-                    self._checkpoint[var] = var.saved_output._ad_create_checkpoint()
+                    if var in global_deps:
+                        self._checkpoint[var] = var._checkpoint
```
Ig-dolci, line 816:
```suggestion 
                        # A new checkpoint object is not necessary here.
                        self._checkpoint[var] = var._checkpoint
```

pyadjoint/checkpointing.py
```diff
@@ -78,9 +84,13 @@ def __init__(self, schedule, tape):
         self.forward_schedule.append(self._current_action)
         # Tell the tape to only checkpoint input data until told otherwise.
         self.tape.latest_checkpoint = 0
-        self.end_timestep(-1)
         self._keep_init_state_in_work = False
         self._adj_deps_cleaned = False
+        self._gc_time_frequency = gc_timestep_frequency
```
Ig-dolci, line 89:
```suggestion
        # The user can manually invoke the garbage collector if Python fails to
        # track and clean all checkpoint objects in memory properly.
        self._gc_time_frequency = gc_timestep_frequency
```

pyadjoint/checkpointing.py
```diff
@@ -361,11 +389,11 @@ def _(self, cp_action, progress_bar, markings, functional=None, **kwargs):
             current_step = self.tape.timesteps[step]
             for block in reversed(current_step):
                 block.evaluate_adj(markings=markings)
-                if not self._adj_deps_cleaned:
+                if not current_step._adj_deps_cleaned:
                     for out in block._outputs:
-                        if not out.marked_in_path:
+                        if not out.marked_in_path and out in current_step.adjoint_dependencies:
```
Ig-dolci, line 394:
```suggestion
                        if not out.marked_in_path:
```

pyadjoint/checkpointing.py
```diff
@@ -78,9 +84,14 @@ def __init__(self, schedule, tape):
         self.forward_schedule.append(self._current_action)
         # Tell the tape to only checkpoint input data until told otherwise.
         self.tape.latest_checkpoint = 0
-        self.end_timestep(-1)
         self._keep_init_state_in_work = False
-        self._adj_deps_cleaned = False
+        # The user can manually invoke the garbage collector if Python fails to
+        # track and clean all checkpoint objects in memory properly.
```
connorjward, line 89: This is confusing because setting `gc_timestep_frequency` suggests that GC is being run automatically, whereas here you say manually

pyadjoint/checkpointing.py
```diff
@@ -164,11 +175,22 @@ def _(self, cp_action, timestep):
             ):
                 for package in self.tape._package_data.values():
                     package.continue_checkpointing()
+            if timestep == 1:
+                # Store the possible global dependencies.
+                for deps in self.tape.timesteps[timestep - 1].checkpointable_state:
+                    self._global_deps.add(deps)
+            else:
+                deps_to_clear = self._global_deps - self._global_deps.intersection(
```
connorjward, line 183: I think you might want `set.difference` https://docs.python.org/3/library/stdtypes.html#frozenset.difference

pyadjoint/checkpointing.py
```diff
@@ -164,11 +175,22 @@ def _(self, cp_action, timestep):
             ):
                 for package in self.tape._package_data.values():
                     package.continue_checkpointing()
+            if timestep == 1:
```
connorjward, line 178: I would want a big comment here explaining what is going on, perhaps even with an example. What are `_global_deps`? Why does removing `_checkpoint` do the right thing?

pyadjoint/checkpointing.py
```diff
@@ -182,6 +204,9 @@ def _(self, cp_action, timestep):
                 # Activate disk checkpointing only in the checkpointing process.
                 for package in self.tape._package_data.values():
                     package.pause_checkpointing()
+
+            if self._gc_timestep_frequency and timestep % self._gc_timestep_frequency == 0:
+                gc.collect(self._gc_generation)
```
connorjward, line 209: I would be tempted to add a `logging.info("Running a garbage collection cycle")` or similar here. Logging is already used inside pyadjoint [here](https://github.com/dolfin-adjoint/pyadjoint/blob/da58d45c5a4dafe529e215f977e108fbabc2f360/pyadjoint/control.py#L50) for example

pyadjoint/checkpointing.py
```diff
@@ -346,6 +371,9 @@ def _(self, cp_action, progress_bar, functional=None, **kwargs):
                         if bv not in current_step.adjoint_dependencies.union(to_keep):
                             bv._checkpoint = None
 
+            if self._gc_timestep_frequency and step % self._gc_timestep_frequency == 0:
+                gc.collect(self._gc_generation)
```
connorjward, line 375: possible logging as above

pyadjoint/checkpointing.py
```diff
@@ -379,6 +407,8 @@ def _(self, cp_action, progress_bar, markings, functional=None, **kwargs):
                         out.reset_variables(("adjoint", "hessian"))
                     if cp_action.clear_adj_deps and out not in to_keep:
                         out._checkpoint = None
+            if self._gc_timestep_frequency and step % self._gc_timestep_frequency == 0:
+                gc.collect(self._gc_generation)
```
connorjward, line 411: logging as above

pyadjoint/tape.py
```diff
@@ -312,12 +312,23 @@ def enable_checkpointing(self, schedule):
         Args:
             schedule (checkpoint_schedules.schedule): A schedule provided by the
             checkpoint_schedules package.
+            gc_timestep_frequency (int): The timestep frequency for garbage collection.
```
connorjward, line 315: Same docstring comment as above.

Also, why is this being repeated? It might make more sense to say: "for more information about this argument refer to :class:`CheckpointManager`"

pyadjoint/tape.py
```diff
@@ -781,25 +792,35 @@ def __init__(self, blocks=()):
         # A dictionary mapping the block variables in the checkpointable state
         # to their checkpoint values.
         self._checkpoint = {}
+        # A flag to indicate whether the adjoint dependencies have been cleaned
+        # from the outputs not marked in the path..
```
connorjward, line 796:
```suggestion
        # from the outputs not marked in the path.
```

pyadjoint/tape.py
```diff
@@ -781,25 +792,35 @@ def __init__(self, blocks=()):
...
+    def checkpoint(self, checkpointable_state, adj_dependencies, global_deps):
         """Store a copy of the checkpoints in the checkpointable state.
 
         Args:
             checkpointable_state (bool): If True, store the checkpointable state
             required to restart from the start of a timestep.
             adj_dependencies (bool): If True, store the adjoint dependencies required
             to compute the adjoint of a timestep.
+            global_deps (set): The set of global dependencies. These dependencies should
+            not be time-dependent and should be in the checkpointable states at every time step.
```
connorjward, line 813: I'd love to have more detail on this.

pyadjoint/tape.py
Ig-dolci: Done.

pyadjoint/checkpointing.py
Ig-dolci: Done.

pyadjoint/checkpointing.py
Ig-dolci: Done.

pyadjoint/checkpointing.py
Ig-dolci: Done

pyadjoint/checkpointing.py
```diff
@@ -37,6 +39,14 @@ class CheckpointManager:
     Args:
         schedule (checkpoint_schedules.schedule): A schedule provided by the `checkpoint_schedules` package.
         tape (Tape): A list of blocks :class:`Block` instances.
+        gc_timestep_frequency (None or int): The number of timesteps between garbage collections. The default
+        is `None`, which means no invoking the garbage collector during the executions. If an integer is
+        provided, the garbage collector is applied every `gc_timestep_frequency` timestep. That is useful when
+        being affected by the Python fails to track and clean all checkpoint objects in memory properly.
```
Ig-dolci, line 45: @connorjward here are the updated docs for `gc_timestep_frequency`.

pyadjoint/checkpointing.py
```diff
@@ -78,9 +88,15 @@ def __init__(self, schedule, tape):
...
         self.tape.latest_checkpoint = 0
-        self.end_timestep(-1)
         self._keep_init_state_in_work = False
-        self._adj_deps_cleaned = False
+        self._gc_timestep_frequency = gc_timestep_frequency
+        self._gc_generation = gc_generation
+        # ``self._global_deps`` stores checkpoint dependencies that remain unchanged across
+        # timesteps (``self.tape.timesteps``). During the forward taping process, the code
+        # checks whether a dependency is in ``self._global_deps`` to avoid unnecessary clearing
+        # and recreation of its checkpoint data.
```
Ig-dolci, line 97: @connorjward Here, I try to explain why ``self._global_deps`` and what is going to store.

pyadjoint/checkpointing.py
```diff
@@ -164,11 +180,29 @@ def _(self, cp_action, timestep):
...
                 for package in self.tape._package_data.values():
                     package.continue_checkpointing()
+            if timestep == 1:
+                # Store the possible global dependencies.
+                for deps in self.tape.timesteps[timestep - 1].checkpointable_state:
+                    self._global_deps.add(deps)
+            else:
+                # Check if the block variables stored at `self._global_deps` are still
+                # dependencies in the previous timestep. If not, will remove them from the
+                # global dependencies.
```
Ig-dolci, line 190: A complement to the docs added when the ``self._global_deps`` was created.

pyadjoint/checkpointing.py
```diff
@@ -180,12 +185,19 @@ def _(self, cp_action, timestep):
                 for deps in self.tape.timesteps[timestep - 1].checkpointable_state:
                     self._global_deps.add(deps)
             else:
-                deps_to_clear = self._global_deps - self._global_deps.intersection(
-                    self.tape.timesteps[timestep - 1].checkpointable_state)
+                # Check if the block variables stored at `self._global_deps` are still
+                # dependencies in the previous timestep. If not, will remove them from the
+                # global dependencies.
+                deps_to_clear = self._global_deps.difference(self._global_deps.intersection(
```
connorjward, line 191: I don't think you need to have the `intersection` here.

pyadjoint/tape.py
```diff
@@ -809,15 +809,16 @@ def checkpoint(self, checkpointable_state, adj_dependencies, global_deps):
...
-            not be time-dependent and should be in the checkpointable states at every time step.
+            global_deps (set): This set stores the common dependencies for all timesteps.
+            For additional information, please refer to the :class:`CheckpointManager`
+            documentation.
         """
         with stop_annotating():
             if checkpointable_state:
                 for var in self.checkpointable_state:
                     if var in global_deps:
                         # Create a new checkpoint object is not necessary here
```
connorjward, line 820:
```suggestion
                        # Creating a new checkpoint object is not necessary here
```

pyadjoint/tape.py
```diff
@@ -809,15 +809,16 @@ def checkpoint(self, checkpointable_state, adj_dependencies, global_deps):
...
-                        # because the global dependencies are not time-dependent.
+                        # because the global dependencies do not change all the time.
```
connorjward, line 821:
```suggestion
                        # because the global dependencies do not change.
```

pyadjoint/checkpointing.py
```diff
@@ -180,12 +185,19 @@ def _(self, cp_action, timestep):
                 for deps in self.tape.timesteps[timestep - 1].checkpointable_state:
                     self._global_deps.add(deps)
             else:
-                deps_to_clear = self._global_deps - self._global_deps.intersection(
-                    self.tape.timesteps[timestep - 1].checkpointable_state)
+                # Check if the block variables stored at `self._global_deps` are still
+                # dependencies in the previous timestep. If not, will remove them from the
+                # global dependencies.
```
connorjward, line 190:
```suggestion
                # Check if the block variables stored in `self._global_deps` are still
                # dependencies in the previous timestep. If not, remove them from the
                # global dependencies.
```

pyadjoint/checkpointing.py
connorjward: But I could be wrong.

pyadjoint/checkpointing.py
```diff
@@ -180,12 +185,19 @@ def _(self, cp_action, timestep):
...
+                deps_to_clear = self._global_deps.difference(self._global_deps.intersection(
+                    self.tape.timesteps[timestep - 1].checkpointable_state))
+
+                # Remove the block variables that are not global dependencies.
+                self._global_deps.difference_update(deps_to_clear)
+
+                # Clear the checkpoint once it is not a global dependency and should be stored
+                # only in the ``self.tape.timesteps`` checkpoints when needed.
```
connorjward, line 198: I'm afraid I don't quite understand what this means. Could you rephrase this?

connorjward (changes requested):

pyadjoint/checkpointing.py
Ig-dolci: Better to understand this text: 
```
For no global dependencies, checkpoint storage occurs at a ``self.tape`` timestep only when required by an action from the schedule. Thus, we have to clear the checkpoint of block variables excluded from the ``self._global_deps``.

```

connorjward: Yeah I think that's good. Thanks.

angus-g:
I think something that could be worth looking at here is the circular reference between `OverloadedType` and `BlockVariable`:
https://github.com/dolfin-adjoint/pyadjoint/blob/c7d73923ef77ca439311b9c044afc7e34580b54f/pyadjoint/overloaded_type.py#L96-L98

In effect, any subclass of `OverloadedType` can *only* be deleted after garbage collection, not reference counting. I think this means that several data arrays sit around for longer than they should, rather than being deleted when their owner goes out of scope.

connorjward:
> In effect, any subclass of OverloadedType can only be deleted after garbage collection, not reference counting. I think this means that several data arrays sit around for longer than they should, rather than being deleted when their owner goes out of scope.

Is there any chance that this could be made a weakref so as to avoid this cycle?

dham:
> > In effect, any subclass of OverloadedType can only be deleted after garbage collection, not reference counting. I think this means that several data arrays sit around for longer than they should, rather than being deleted when their owner goes out of scope.
> 
> Is there any chance that this could be made a weakref so as to avoid this cycle?

I think so. I think the BlockVariable should not prolong the lifetime of the Overloaded Type, so `BlockVariable.output` should be a weakref. We have to be careful that we're not abusing `Blockvariable.output` anywhere (i.e. relying on it as a source of information after the operation has been taped).

Ig-dolci:
> > In effect, any subclass of OverloadedType can only be deleted after garbage collection, not reference counting. I think this means that several data arrays sit around for longer than they should, rather than being deleted when their owner goes out of scope.
> 
> Is there any chance that this could be made a weakref so as to avoid this cycle?

I already tried weakref for `BlockVariables.output`, but I hit a number of errors I do not remember now. It is a very careful work to do.

angus-g:
I had a hacky go at that before, which worked for the forward run of the tape (with some implementation ugliness). The underlying OverloadedVarable was deleted at some point before/during the adjoint call, so that might need some care.

jrmaddison:
Looks easier to break the cycle on the other side, see #194 for an attempt.

Ig-dolci:
To make you updated:

I have tested this PR merged to the [PR 194](https://github.com/dolfin-adjoint/pyadjoint/pull/194)  against the [PR 194](https://github.com/dolfin-adjoint/pyadjoint/pull/194)  (**only**) for [Burgers' equation](https://github.com/dolfin-adjoint/pyadjoint/blob/master/tests/firedrake_adjoint/test_burgers_newton.py#L122) using the following setup: 40,000 DoFs and 1,000 time steps.  

The chart below uses `SingleDiskStorageSchedule`. The black line represents the results related to the [PR 194](https://github.com/dolfin-adjoint/pyadjoint/pull/194)  (**only**), and the blue line represents this PR merged to the [PR 194](https://github.com/dolfin-adjoint/pyadjoint/pull/194)  using `gc_timestep_frequency=100`.
mem_gc

Ig-dolci:
I will also check the [PR 4033](https://github.com/firedrakeproject/firedrake/pull/4033) using the same example and add it here.

Ig-dolci:
Now using  [firedrake PR 4033](https://github.com/firedrakeproject/firedrake/pull/4033) merged to [firedrake PR 4020](https://github.com/firedrakeproject/firedrake/pull/4020) that automatically uses  `SingleDiskStorageSchedule`.

Again, I have tested this PR merged to the [pyadjoint PR 194](https://github.com/dolfin-adjoint/pyadjoint/pull/194)  against the [PR 194](https://github.com/dolfin-adjoint/pyadjoint/pull/194)  (**only**) for [Burgers' equation](https://github.com/dolfin-adjoint/pyadjoint/blob/master/tests/firedrake_adjoint/test_burgers_newton.py#L122) using the following setup: 40,000 DoFs and 1,000 time steps.  

The black line represents the results related to the [PR 194](https://github.com/dolfin-adjoint/pyadjoint/pull/194)  (**only**), and the blue line represents this PR merged to the [PR 194](https://github.com/dolfin-adjoint/pyadjoint/pull/194)  using `gc_timestep_frequency=100`.
final

pyadjoint/checkpointing.py
Ig-dolci, line 45:
```suggestion
        gc_timestep_frequency (None or int): The number of timesteps between additional garbage collections. The default
        is `None`, which means no additional garbage collection. If an integer is
        provided, the garbage collector is applied every `gc_timestep_frequency` timesteps. This is useful if delays
        in object collection are causing excessive memory usage.
```

pyadjoint/checkpointing.py
Ig-dolci: I will check this.

pyadjoint/tape.py
```diff
@@ -313,12 +313,23 @@ def enable_checkpointing(self, schedule):
...
+            gc_generation (int): The generation for garbage collection. For additional
+            information, please refer to the :class:`CheckpointManager` documentation.
         """
         if self._blocks:
             raise CheckpointError(
                 "Checkpointing must be enabled before any blocks are added to the tape."
             )
-        self._checkpoint_manager = CheckpointManager(schedule, self)
+
+        if gc_timestep_frequency and not isinstance(gc_timestep_frequency, int):
```
Ig-dolci, line 327:
```suggestion
        if gc_timestep_frequency is not None and not isinstance(gc_timestep_frequency, int):
```

connorjward (approved):
Looks great. Very readable now.

Issue #169, opened 2024-09-24
Checkpointing and referencing of variables


jrmaddison:
The pyadjoint tape references backend variables. This means that any memory allocated for the forward variables, during the forward calculation, is referenced by the tape. This can prevent memory usage being reduced by checkpointing.

**Example**
```
from firedrake import *
from firedrake.adjoint import *
from checkpoint_schedules import MultistageCheckpointSchedule

N = 100

mesh = UnitIntervalMesh(1)
space = FunctionSpace(mesh, "Lagrange", 1)

tape = get_working_tape()
tape.enable_checkpointing(MultistageCheckpointSchedule(N, 3, 0))

u = Function(space, name="u").interpolate(Constant(2.0))
continue_annotation()
for _ in tape.timestepper(iter(range(N))):
    u_ = Function(space, name="u")
    assemble(Interpolate(u + u, space), tensor=u_)
    u = u_
    del u_
pause_annotation()
del u

deps = set()
outputs = set()
for block in tape._blocks:
    for dep in block.get_dependencies():
        if isinstance(dep.output, Function):
            deps.add(dep.output.count())
    for dep in block.get_outputs():
        if isinstance(dep.output, Function):
            outputs.add(dep.output.count())

print(f"{len(deps)=}")
print(f"{len(outputs)=}")
```
leads to output
```
len(deps)=100
len(outputs)=100
```

PR #194, opened 2025-02-11
Break `OvertypedType` -> `BlockVariable` -> `OverloadedType` reference cycle

jrmaddison:
Use a `weakref` for `OverloadedType.block_variable`, accessed indirectly via a `property`. The result persists for as long as it is used elsewhere (e.g. on the tape, referenced by a `Control`, etc).

Example use:
```
ot.create_block_variable()
object_0._block_variable = ot.block_variable
object_1._block_variable = ot.block_variable
```
First line creates and then immediately destroys a `BlockVariable`. Second line creates another `BlockVariable`, holding a reference to it, so that the third line references the same `BlockVariable`. Net effect: attach a new `BlockVariable` to `ot`, hold two references on `object_0` and `object_1`, hold a weak reference on `ot`.

Could be optimized by adding an `OverloadedType.clear_block_variable` method to replace the first line of the above example

```
def clear_block_variable(self):
    self._block_variable = lambda: None
```

which avoids the creation and immediate destruction of a `BlockVariable`.

### DIFF
--- a/pyadjoint/overloaded_type.py
+++ b/pyadjoint/overloaded_type.py
@@ -1,3 +1,4 @@
+import weakref
 from .block_variable import BlockVariable
 from .tape import get_working_tape
 
@@ -74,7 +75,7 @@ class OverloadedType(object):
     """
 
     def __init__(self, *args, **kwargs):
-        self.block_variable = None
+        self._block_variable = lambda: None
         self.create_block_variable()
 
     @classmethod
@@ -93,9 +94,18 @@ class OverloadedType(object):
         """
         return cls(obj)
 
+    @property
+    def block_variable(self):
+        block_variable = self._block_variable()
+        return self.create_block_variable() if block_variable is None else block_variable
+
+    @block_variable.setter
+    def block_variable(self, value):
+        self._block_variable = weakref.ref(value)
+
     def create_block_variable(self):
-        self.block_variable = BlockVariable(self)
-        return self.block_variable
+        self.block_variable = block_variable = BlockVariable(self)
+        return block_variable
 
     def _ad_convert_type(self, value, options={}):
         """This method must be overridden.
--- a/pyadjoint/overloaded_type.py
+++ b/pyadjoint/overloaded_type.py
@@ -76,7 +76,6 @@ class OverloadedType(object):
 
     def __init__(self, *args, **kwargs):
         self._block_variable = lambda: None
-        self.create_block_variable()
 
     @classmethod
     def _ad_init_object(cls, obj):
--- a/pyadjoint/overloaded_type.py
+++ b/pyadjoint/overloaded_type.py
@@ -65,6 +65,34 @@ def register_overloaded_type(overloaded_type, classes=None):
     return overloaded_type
 
 
+class Weakref:
+    """Weakref which is picklable if the referenced obj is picklable.
+    """
+
+    def __init__(self, obj=None):
+        self._init(obj)
+
+    def _init(self, obj):
+        if obj is None:
+            self._obj = lambda: None
+        else:
+            self._obj = weakref.ref(obj)
+
+    def __call__(self):
+        return self._obj()
+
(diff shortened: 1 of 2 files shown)


Ig-dolci:
I have tested this PR against the master branch for [Burgers' equation](https://github.com/dolfin-adjoint/pyadjoint/blob/master/tests/firedrake_adjoint/test_burgers_newton.py#L122) using the following setup: 40,000 DoFs and 1,000 time steps.  

The first chart below shows results **without checkpointing**. The black line represents the master branch, and the blue line corresponds to the current PR:  
mem_no_chk  

The second chart shows results using the `SingleMemoryStorageSchedule`. Again, the black line represents the master branch, and the blue line represents the current PR:  
mem_chk

jrmaddison:
Were the patches in https://github.com/firedrakeproject/firedrake/pull/4033 also applied? Otherwise reference cycles might exist elsewhere.

Ig-dolci:
> Were the patches in [firedrakeproject/firedrake#4033](https://github.com/firedrakeproject/firedrake/pull/4033) also applied? Otherwise reference cycles might exist elsewhere.

Not yet. Will be my next step. I will add here.

Ig-dolci:
Now using the firedrake branch from [PR 4033](https://github.com/firedrakeproject/firedrake/pull/4033):

This PR against the pyadjoint  master branch for [Burgers' equation](https://github.com/dolfin-adjoint/pyadjoint/blob/master/tests/firedrake_adjoint/test_burgers_newton.py#L122) using the following setup: 40,000 DoFs and 1,000 time steps.  

The first chart below shows results **without checkpointing**. The black line represents the master branch, and the blue line corresponds to the current PR:  
mem_no_chk_angus


The second chart shows results using the `SingleMemoryStorageSchedule`. Again, the black line represents the pyadjoint  master branch, and the blue line represents the current PR:  
mem_chk_angus

dham:
Can this please be checked for safety in the case of `with stop_annotating(modifies=...)`. I think that case causes new block variables to be created that aren't (yet) referenced elsewhere. I am nervous that there might be other similar cases.

jrmaddison:
> Can this please be checked for safety in the case of `with stop_annotating(modifies=...)`. I think that case causes new block variables to be created that aren't (yet) referenced elsewhere.

`create_block_variable` in `stop_annotating.__exit__`  leads to elements of `modifies` having new block variables attached, which are then immediately destroyed. This correctly breaks the link to any previously attached block variable. A later access to `modifies_element.block_variable` will create another new block variable, but that's fine, it's still different from the one that was attached before the `stop_annotating` block.

> I am nervous that there might be other similar cases.

The pattern
```
ot.block_variable.tlm_value = tlm_value
continue_annotation()
...
pause_annotating()
```
will lead to the `tlm_value` being dropped, since the block variable is created and then immediately destroyed. That's a change in behaviour, but I'm not sure it makes sense to access a block variable before it's annotated on a tape.

dham:
> > Can this please be checked for safety in the case of `with stop_annotating(modifies=...)`. I think that case causes new block variables to be created that aren't (yet) referenced elsewhere.
> 
> `create_block_variable` in `stop_annotating.__exit__` leads to elements of `modifies` having new block variables attached, which are then immediately destroyed. This correctly breaks the link to any previously attached block variable. A later access to `modifies_element.block_variable` will create another new block variable, but that's fine, it's still different from the one that was attached before the `stop_annotating` block.

Would it be better for that operation to instead just remove the block variables then?

jrmaddison:
> Would it be better for that operation to instead just remove the block variables then?

That's easy to add (the `OverloadedType.clear_block_variable` method mentioned above).

Note that this PR doesn't address #169, which requires the `OverloadedType` to never hold (or else eventually drop) references to variables. Breaking the cycle on the `OverloadedType` side could fix both #169 and this reference cycle -- it just looks much harder to do.

dham (approved):

PR #197, opened 2025-02-13
Convert TimeStep block variables to lists

angus-g:
Although the sets of block variables owned by TimeStep are usually used as sets, their order is not deterministic. In single-process contexts, this doesn't make any practical difference. However, when using MPI all the ranks must collectively checkpoint functions. In practice, I think these lists are sufficiently small that they can be converted to sets as-needed, while crucially retaining their insertion order.


Ig-dolci:
Thank you. This PR fixes a test error I was planning to track today.

angus-g:
Ah great! Upon reflection, it could be more elegant to simply sort the sets when it's time to checkpoint. I'm not sure what a stable sort key would be on a block variable, however.

angus-g:
> I am a little nervous, though about the duplicates in the list.

Agreed, I think as-is this is likely to introduce errors. Maybe it's possible that multiple blocks have the same adjoint dependency within a timestep, or something like that? There must be a fairly simple to construct pattern to test this... 

> Could the state ever get big enough that that's a problem?

I suppose the ideal pattern is to have many timesteps but with each timestep fairly constrained. Of course, it would be easy enough to blow out to quadratic performance which would not be ideal!

> One way to do this would be to have a set subclass which also records a list in insertion order

(thinking aloud) I suppose de-duplicating on insertion achieves the same outcome, although if there are a lot (a lot!) of state/adjoint variables the linear scans add up? There's also the slight difference in deletion semantics: `list.remove` errors when the element isn't found, but `set.discard` always succeeds. Seems a little fiddly either keeping a list, or using a special set class! Easier to reason about correctness/performance by keeping it as a set though.

dham:
> (thinking aloud) I suppose de-duplicating on insertion achieves the same outcome, although if there are a lot (a lot!) of state/adjoint variables the linear scans add up? There's also the slight difference in deletion semantics: `list.remove` errors when the element isn't found, but `set.discard` always succeeds. Seems a little fiddly either keeping a list, or using a special set class! Easier to reason about correctness/performance by keeping it as a set though.

I think that the set subclass would internally also store a list. The scan on insert would be `item in self` which uses the set membership test which is O(1). Similarly, the `discard` implementation would be:
```
def discard(self, item):
    if item in self:
        self._list.remove(item)
    super().discard(item)
```
That would have the same semantics as plain set discard.

pyadjoint/ordered_set.py
```diff
@@ -0,0 +1,134 @@
+from typing import Hashable, Iterable
+
+
+class OrderedSet(set):
```
jrmaddison, line 4: Seems more natural to inherit from `collections.abc.MutableSet`.

pyadjoint/ordered_set.py
```diff
@@ -0,0 +1,134 @@
...
+        self._elements.append(obj)
+
+    def remove(self, obj: Hashable):
+        """Remove obj from this set, it must be a member."""
+
+        if obj not in self:
+            raise KeyError(obj)
+
+        super().remove(obj)
+        self._elements.remove(obj)
```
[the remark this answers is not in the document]
angus-g: Yeah, good point. Now it's just backed by a dictionary.

pyadjoint/ordered_set.py
angus-g: Ah yeah, I always forget about those. Especially with dictionary backing, this seems more reasonable.

dham (approved):

PR #207, opened 2025-04-09
Bump version

connorjward:
Necessary for https://github.com/dolfin-adjoint/pyadjoint/issues/205

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pyadjoint-ad"
-version = "2023.0.0"
+version = "2025.04.0"
 dependencies = [
   "checkpoint_schedules",
   "scipy>=1.0",


jorgensd (approved):
Is the idea to use months as the second counter?
If so, should one go for `YYYY-MM-DD` in case there are multiple releases within the same month, i.e. Calendar Versioning: https://en.wikipedia.org/wiki/Software_versioning

connorjward:
> Is the idea to use months as the second counter? If so, should one go for `YYYY-MM-DD` in case there are multiple releases within the same month, i.e. Calendar Versioning: https://en.wikipedia.org/wiki/Software_versioning

We are planning on doing <year>.<month>.<patch> like Ubuntu.

jorgensd:
> > Is the idea to use months as the second counter? If so, should one go for `YYYY-MM-DD` in case there are multiple releases within the same month, i.e. Calendar Versioning: https://en.wikipedia.org/wiki/Software_versioning
> 
> We are planning on doing `<year>.<month>.<patch>` like Ubuntu.

Ok

francesco-ballarin:
Just two FYIs: 
- the commit message says that the patch version was supposed to be `01`, while the actual version turned out to have `0` instead of `01`.
- it seems to me that PyPI hides the leading 0 in `04`, see https://pypi.org/project/pyadjoint-ad/#history . If that is the case, maybe in future releases you may want to hide it from `pyproject.toml` as well. `2025.4` and `2025.04` will reasonably be equivalent version numbers, but I can see this being a source of confusion for some users...

PR #210, opened 2025-06-25
mark functional dependencies separately from control dependents

### LINKED ISSUE #209: Incorrect derivatives when using SingleMemoryStorageSchedule with control-independent assignments
We have encountered [an issue](https://github.com/g-adopt/g-adopt/issues/237) where adjoint derivatives are incorrect when using SingleMemoryStorageSchedule, but correct with no checkpointing. 

Consider [this reproducer](https://gist.github.com/sghelichkhani/cac378b9bf649359465b9e439bf04e59).  When running 

```
python script.py none  # Running the equivalent of a Taylor test
```
you would get:

```
using scheduler: NoneType
        dJdm 1: 5.537600874557515
        reduced functional: 4.839999999999989
        dJdm 2: 6.091360962013265
        reduced functional: 4.839999999999989
        dJdm 3: 6.091360962013265
        eps: 0.01, res: 0.00019357364048094156
        eps: 0.005, res: 4.839341012216267e-05
        eps: 0.0025, res: 1.209835253416884e-05
        eps: 0.00125, res: 3.024588136022431e-06
```
Now using a memory scheduler: 
```
python script.py none    # Not using any scheduler
```
we get: 

```
using scheduler: SingleMemoryStorageSchedule
        dJdm 1: 5.537600874557515
        reduced functional: 4.839999999999989
        dJdm 2: 3.0456804810066327
        reduced functional: 4.839999999999989
        dJdm 3: 3.0456804810066327
        eps: 0.01, res: 0.03065037845054727
        eps: 0.005, res: 0.015276795815155326
        eps: 0.0025, res: 0.007626299555050751
        eps: 0.00125, res: 0.0038101251893943133
```

## Cause

The problem arises from this code block:

https://github.com/dolfin-adjoint/pyadjoint/blob/a4f940a7899809a65d43d9fb9e6049a70755a30a/pyadjoint/checkpointing.py#L407-L417

This pruning logic assumes that any block output not marked is irrelevant for future reverse passes. But the assumption fails if the block (e.g. `FunctionAssignBlock`) is control-independent and its output is used later in the graph. Since it's unmarked, its output gets discarded from `adjoint_dependencies`, and with `SingleMemoryStorageSchedule` clears its checkpoint. 

On the next reverse pass, the value is reconstructed incorrectly, and the derivative is wrong. In a way we seem to be too eager to discard checkpoints.

dham:
Currently the marking of block variables on the tape is only on the basis of whether they depend on the control. For computing an adjoint dependency, that's not the right test. This PR also marks controls that are dependencies of the functional.

Closes #209

### DIFF
--- a/pyadjoint/tape.py
+++ b/pyadjoint/tape.py
@@ -539,7 +539,7 @@ class Tape(object):
             node.is_functional_dependency = True
         try:
             yield
-        except:
+        finally:
             for node in nodes:
                 node.is_functional_dependency = False
  


dham:
Firedrake tests running at: https://github.com/firedrakeproject/firedrake/pull/4403

dham (commented):
Just needs the MFE ported in as a test.

PR #212, opened 2025-06-30
Factor the TLM calculation out of Hessian calculation

JHopeCollins:
Currently there is no way of asking pyadjoint to calculate the action of the TLM unless it's as part of the Hessian action.
This PR just make the TLM action a first class citizen like the adjoint and Hessian, with a driver and a `ReducedFunctional` method of its own.

### DIFF
--- a/pyadjoint/drivers.py
+++ b/pyadjoint/drivers.py
@@ -81,7 +81,7 @@ if deprecated is not None:
     )(compute_gradient)
 
 
-def compute_hessian(J, m, m_dot, tape=None, apply_riesz=False):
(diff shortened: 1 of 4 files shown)


dham (commented):
Looks good to me, but will need tests (probably in Firedrake)

jorgensd:
> Looks good to me, but will need tests (probably in Firedrake)

I like this as well. I can add a test to https://github.com/jorgensd/dxa/blob/main/tests/test_poisson_mother.py#L90 as well, as I already compute the Tlm by hand there.

PR #213, opened 2025-07-01
Python PETSc Mat for ReducedFunctional

JHopeCollins:
This PR introduces a `ReducedFunctionalMat`, which is a Python type PETSc.Mat which applies the action of a `ReducedFunctional`.

By default it applies the action of the Hessian, which is needed for TAO's Newton optimiser.
If `Jhat.functional` is not a scalar then it can also apply the action of the TLM or adjoint models. Examples of when these would be needed include 1) constructing saddle point formulations of an optimisation problem, 2) constructing preconditioners for the Hessian with a certain structure, 3) calculating the eigendecomposition of a reduced functional (https://github.com/firedrakeproject/firedrake/pull/4529).

Firedrake tests: https://github.com/firedrakeproject/firedrake/pull/4669

### DIFF
--- a/pyadjoint/optimization/tao_solver.py
+++ b/pyadjoint/optimization/tao_solver.py
@@ -181,6 +181,11 @@ class ReducedFunctionalMatBase:
     Adjoint : U* -> V*
     Hessian : V x U* -> V* | V -> V*
 
+    Child classes must implement:
+    - mult_impl
+    - multHermitian_impl
+    - update_adjoint
+
     Args:
         rf (ReducedFunctional): Defines the forward model. Used to compute Mat actions.
         apply_riesz (bool): Whether to apply the riesz map before returning the
@@ -189,9 +194,6 @@ class ReducedFunctionalMatBase:
         always_update_tape (bool): Whether to force reevaluation of the forward model
             every time `mult` is called. If needs_adjoint_update then this will also force
             the adjoint model to be reevaluated at every call to `mult`.
-        update_adjoint (bool): Whether to update the adjoint as well as the forward
-            model when updating the Mat. If True then `rf.derivative` will be called by
-            the `update` method. Required for Hessian but not for TLM or Adjoint Mats.
         needs_functional_interface: Whether to create a PETScVecInterface for rf.functional.
         comm (Optional[petsc4py.PETSc.Comm,mpi4py.MPI.Comm]): Communicator that the rf is defined over.
     """
@@ -199,7 +201,6 @@ class ReducedFunctionalMatBase:
     def __init__(self, rf, *,
                  apply_riesz=False, appctx=None,
                  always_update_tape=False,
-                 update_adjoint=False,
                  needs_functional_interface=False,
                  comm=PETSc.COMM_WORLD):
         comm = valid_comm(comm)
@@ -215,7 +216,6 @@ class ReducedFunctionalMatBase:
             self.functional_interface = PETScVecInterface(
                 rf.functional, comm=comm)
 
-        self.update_adjoint = update_adjoint
         self._m = new_control_variable(rf)
         self._shift = 0
         self.always_update_tape = always_update_tape
@@ -224,7 +224,7 @@ class ReducedFunctionalMatBase:
     def update(cls, obj, x, A, P):
         ctx = A.getPythonContext()
         ctx.control_interface.from_petsc(x, ctx._m)
-        ctx.update_tape_values(update_adjoint=ctx.update_adjoint)
+        ctx.update_tape_values(update_adjoint=cls.update_adjoint())
         ctx._shift = 0
 
     def shift(self, A, alpha):
@@ -279,6 +279,14 @@ class ReducedFunctionalMatBase:
         raise NotImplementedError(
             "Must provide implementation of the Hermitian action of this matrix on an OverloadedType")
 
+    @classmethod
+    def update_adjoint(self):
+        """
+        Whether to update the adjoint as well as the forward model when updating
+        the Mat. If True then `rf.derivative` will be called by the `update` method.
+        """
+        raise NotImplementedError
+
 
 class ReducedFunctionalHessianMat(ReducedFunctionalMatBase):
     """
@@ -302,12 +310,16 @@ class ReducedFunctionalHessianMat(ReducedFunctionalMatBase):
                  always_update_tape=False, comm=PETSc.COMM_WORLD):
 
         super().__init__(rf, apply_riesz=apply_riesz, appctx=appctx,
-                         needs_functional_interface=False, update_adjoint=True,
+                         needs_functional_interface=False,
                          always_update_tape=always_update_tape, comm=comm)
 
         self.xinterface = self.control_interface
         self.yinterface = self.control_interface
         self.x = new_control_variable(rf)
+
+    @classmethod
+    def update_adjoint(self):
+        return True
 
     def mult_impl(self, A, x):
         if self.always_update_tape:
@@ -341,12 +353,16 @@ class ReducedFunctionalAdjointMat(ReducedFunctionalMatBase):
                  always_update_tape=False, comm=PETSc.COMM_WORLD):
 
         super().__init__(rf, apply_riesz=apply_riesz, appctx=appctx,
-                         needs_functional_interface=True, update_adjoint=False,
+                         needs_functional_interface=True,
                          always_update_tape=always_update_tape, comm=comm)
 
         self.xinterface = self.functional_interface
         self.yinterface = self.control_interface
         self.x = rf.functional._ad_copy()
+
+    @classmethod
+    def update_adjoint(self):
+        return False
 
     def mult_impl(self, A, x):
         if self.always_update_tape:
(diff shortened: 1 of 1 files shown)


pyadjoint/optimization/tao_solver.py
```diff
@@ -184,35 +472,12 @@ def hessian(self, m, m_dot):
...
         m_dot = Enlist(m_dot)
-        _ = self.reduced_functional(tuple(m_i._ad_copy() for m_i in m))
-        _ = self.reduced_functional.derivative()
+        if self.always_update_tape:
+            _ = self.reduced_functional(tuple(m_i._ad_copy() for m_i in m))
+            _ = self.reduced_functional.derivative()
         ddJ = self.reduced_functional.hessian(tuple(m_dot_i._ad_copy() for m_dot_i in m_dot))
         return m.delist(ddJ)
 
-    def new_control_variable(self):
```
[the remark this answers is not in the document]
JHopeCollins: I made this change because there was somewhere (I forget exactly where) that I wanted to call this method where it didn't otherwise make sense to have one of these objects, and creating the new variables only depends the `ReducedFunctional` anyway.

Do you use this part of the current API? It's not technically part of the public API via `__all__`, but Hyrum's law and all that. If you do use it would you be opposed to this change?

pyadjoint/optimization/tao_solver.py
```diff
@@ -184,35 +472,12 @@ def hessian(self, m, m_dot):
...
-
-        Returns:
-            tuple[OverloadedType]: New variables suitable for storing a control
-                value.
-        """
-
-        return tuple(control._ad_init_zero(dual=False)
-                     for control in self.reduced_functional.controls)
-
-    def new_dual_control_variable(self):
```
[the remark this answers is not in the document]
JHopeCollins: See comment above.

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +130,315 @@ def to_petsc(self, x, y):
...
+    Jhat : V -> U
+    TLM : V -> U
+    Adjoint : U* -> V*
+    Hessian : V x U* -> V* | V -> V*
+
+    Args:
+        rf (ReducedFunctional): Defines the forward model, and used to compute operator  actions.
+        action (RFAction): Whether to apply the TLM, adjoint, or Hessian action.
+        apply_riesz (bool): Whether to apply the riesz map before returning the
+            result of the action to PETSc.
```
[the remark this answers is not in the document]
JHopeCollins: Now each action type is a different base class, `ReducedFunctionalTLMMat` doesn't take the `apply_riesz` argument.

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,368 @@ def to_petsc(self, x, y):
...
+    Args:
+        rf (ReducedFunctional): Defines the forward model, and used to compute operator actions.
+        action (RFAction): Whether to apply the TLM, adjoint, or Hessian action.
+        apply_riesz (bool): Whether to apply the riesz map before returning the
+            result of the action to PETSc.
+        appctx (Optional[dict]): User provided context.
+        always_update_tape (bool): Whether to force reevaluation of the forward model every time
+            `mult` is called. If action is HESSIAN then this will also force the adjoint model to
+            be reevaluated at every call to `mult`.
+        needs_functional_interface: Whether to create a PETScVecInterface for the rf.functional.
```
dham: a vec is required when.....

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,368 @@ def to_petsc(self, x, y):
...
+        if update_adjoint:
+            _ = self.rf.derivative(apply_riesz=False)
+
+    def mult(self, A, x, y):
+        self.xinterface.from_petsc(x, self.x)
+        out = self.mult_impl(A, self.x)
+        self.yinterface.to_petsc(y, out)
+
+        if self._shift != 0:
+            y.axpy(self._shift, x)
```
dham: add hermitian multiply.

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,368 @@ def to_petsc(self, x, y):
...
+            always_update_tape=always_update_tape, comm=comm)
+
+    elif action == RFAction.TLM:
+        ctx = ReducedFunctionalTLMMat(
+            rf, appctx=appctx,
+            always_update_tape=always_update_tape, comm=comm)
+
+    else:
+        raise ValueError(
+            'Unrecognised RFAction: {action}.')
```
dham:
```suggestion
            f'Unrecognised RFAction: {action}.')
```

dham (changes requested):

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,410 @@ def to_petsc(self, x, y):
...
+
+    Args:
+        reduced_functional (ReducedFunctional): The reduced functional whose
+        controls are to be copied.
+        dual (bool): whether to return a dual type. If False then a primal type is returned.
+
+    Returns:
+        tuple[OverloadedType]: New variables suitable for storing a control value.
+    """
+
```
connorjward:
```suggestion
```

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,410 @@ def to_petsc(self, x, y):
...
+class ReducedFunctionalMatBase:
+    """
+    PETSc.Mat Python context to apply the action of a pyadjoint.ReducedFunctional.
+
+    If V is the control space and U is the functional space, each action has the following map:
+    Jhat : V -> U
+    TLM : V -> U
+    Adjoint : U* -> V*
+    Hessian : V x U* -> V* | V -> V*
```
connorjward: Just for me, what does this notation mean?

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,410 @@ def to_petsc(self, x, y):
...
+
+    Args:
+        rf (ReducedFunctional): Defines the forward model, and used to compute operator actions.
+        action (RFOperation): Whether to apply the TLM, adjoint, or Hessian action.
```
connorjward: Wasn't this discussed in the meeting and this was determined a confusing name?

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,410 @@ def to_petsc(self, x, y):
...
+        if self._shift != 0:
+            y.axpy(self._shift, x)
+
+    def mult_impl(self, A, x):
+        """
+        This method must be overriden.
+
+        Provides the implementation of a particular type of ReducedFunctional action.
+        """
+        raise NotImplementedError
```
connorjward: Should have a helpful error message

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,410 @@ def to_petsc(self, x, y):
...
+        ((nrow, Nrow), (ncol, Ncol)),
+        ctx, comm=ctx.control_interface.comm)
+    if action == RFOperation.HESSIAN:
+        mat.setOption(PETSc.Mat.Option.SYMMETRIC, True)
+    mat.setUp()
+    mat.assemble()
+    return mat
+
+
+class RieszMapMatCtx:
```
connorjward: Docstring?

pyadjoint/optimization/tao_solver.py
connorjward: Yes but the option is still called `action`. Though see @jrmaddison's comment that this can probably just go away.

JHopeCollins: See response to that comment.

pyadjoint/optimization/tao_solver.py
```diff
@@ -131,26 +129,410 @@ def to_petsc(self, x, y):
...
+        self._m = new_control_variable(rf)
+        self._shift = 0
+        self.always_update_tape = always_update_tape
+
+    @classmethod
+    def update(cls, obj, x, A, P):
+        ctx = A.getPythonContext()
+        ctx.control_interface.from_petsc(x, ctx._m)
+        ctx.update_tape_values(
+            update_adjoint=(ctx.action == RFOperation.HESSIAN))
```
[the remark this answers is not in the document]
connorjward: I think what @jrmaddison means is we could replace this line with
```py
update_adjoint=isinstance(ctx, ReducedFunctionalHessianMat)
```
then we don't have to have an extra argument.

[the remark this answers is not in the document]
JHopeCollins: As I said above, I don't like that version because its bad practice to require the parent class to know about child classes. It's better OO practice to have the child class provide the information to the parent. This is (almost) one of James' suggested fixes: "or a call to a subclass method which decides whether to update the adjoint".

If there are other child classes in the future then having `update_adjoint` provided means that the parent class won't need changing. For example a Gauss Newton Hessian approximation. Open to extension but closed to modification and all that.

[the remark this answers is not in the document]
connorjward: It's a bit confusing because of dealing with`ctx` instead of `self`.
I think the right solution here is to have `update_adjoint` be a class-level property (it could be defined as an abstract one by the parent class).

PR #214, opened 2025-07-03
Use petsctools

jrmaddison:


connorjward (approved):
This is very pleasing. I think that CI is failing because petsctools is in `master` not `release` which is what we have for our containers. We therefore would need to make a major release of Firedrake, ~which is probably a good thing to do anyway~ (no it isn't).

JHopeCollins:
Is there any reason this shouldn't be merged yet?

connorjward:
> Is there any reason this shouldn't be merged yet?

As I stated above this doesn't work using our current containers. One could (and probably should, as per our other discussion) add petsctools as an optional dependency to pyadjoint.

JHopeCollins:
> > Is there any reason this shouldn't be merged yet?
> 
> As I stated above this doesn't work using our current containers. One could (and probably should, as per our other discussion) add petsctools as an optional dependency to pyadjoint.

I forgot that the most recent Firedrake release doesn't have petsctools in, for some reason I thought that 2025.4.2 included it.
So if `petsctools` is added as an optional dependency in this PR then - if the `pip install` instruction in the CI is also updated - it should pass CI and we can merge?

jrmaddison:
cherry-pick to the rescue. No idea why that commit was dropped.

JHopeCollins:
That "+9,-201" lines changed looks much better, thanks!

PR #216, opened 2025-07-04
Add docstring entries for the `adj_value` kwargs for drivers

jwallwork23:
I noticed that the `compute_gradient` and `solve_adjoint` driver functions have a kwarg `adj_value`, but do not mention this in their docstrings.


JHopeCollins:
This about to be remedied in https://github.com/dolfin-adjoint/pyadjoint/pull/156

PR #223, opened 2025-09-16
Convert ROL interface to use AbstractReducedFunctional

angus-g:
We recover the original behaviour by applying the Riesz map when returning the derivative or Hessian actions. Maybe it would be better to provide a `.dual()` method on the vector and let ROL decide which representation it wants? Probably also needs a test (e.g. on the Firedrake side)?


angus-g:
I think this should be good to go (at least enough to restore functionality for us), I don't know if it should wait for https://github.com/firedrakeproject/firedrake/pull/4581 or not?

PR #225, opened 2025-09-16
Update Docker image for test-firedrake job

jorgensd:
Resolve https://github.com/dolfin-adjoint/pyadjoint/issues/224

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ version: 2
 jobs:
   test-firedrake:
     docker:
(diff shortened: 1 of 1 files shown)


angus-g:
Is that just `|| [ $? == 5 ]` after all the `pytest` commands or is there a slightly cleaner way of doing it?

connorjward:
> Is that just `|| [ $? == 5 ]` after all the `pytest` commands or is there a slightly cleaner way of doing it?

That's probably the best approach.

PR #230, opened 2025-10-09
SciPy Riesz map fix

### LINKED ISSUE #153: `derivative_options` argument to `minimize` is likely incorrect
e.g. for SciPy's L-BFGS-B using a (non $l^2$) Riesz map will lead to the derivative being supplied to the optimizer being inconsistent. SciPy requires the (non-mapped) derivative.

jrmaddison:
#221 did not in fact fix #153, as `ReducedFunctionalNumpy.derivative` and `ReducedFunctionalNumpy.hessian` use a default `apply_riesz=True` (which differs from the `AbstractReducedFunctional` base class).

A second attempt.

### DIFF
--- a/pyadjoint/reduced_functional_numpy.py
+++ b/pyadjoint/reduced_functional_numpy.py
@@ -59,7 +59,7 @@ class ReducedFunctionalNumPy(AbstractReducedFunctional):
         return numpy.array(m_global, dtype="d")
 
     @no_annotations
-    def derivative(self, adj_input=1.0, apply_riesz=True):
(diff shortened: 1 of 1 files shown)


JHopeCollins:
@jrmaddison The H1 riesz map test that is erroneously passing has an `rtol=1e-2` in the `assert_allclose`. This may not be tight enough to pick up on the difference in the "optimised" solution between the different Riesz maps. Can you try tightening up that tolerance? If that fixes the xpass then we can update the firedrake test.

PR #234, opened 2025-10-23
Ignore `FutureWarning` when testing `compute_gradient`

JHopeCollins:
Also adds the newer `compute_derivative` and `compute_tlm` drivers to the top level API with `compute_gradient` and `compute_hessian`.

### DIFF
--- a/tests/pyadjoint/test_enlisting.py
+++ b/tests/pyadjoint/test_enlisting.py
@@ -1,5 +1,8 @@
 from pyadjoint import *
+import pytest
 
+
+@pytest.mark.filterwarnings("ignore:compute_gradient:FutureWarning")
 def test_compute_gradient():
     a = AdjFloat(1.0)
     b = AdjFloat(2.0)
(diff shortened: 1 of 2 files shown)


tests/pyadjoint/test_enlisting.py
```diff
@@ -10,6 +13,16 @@ def test_compute_gradient():
...
 
 
+def test_compute_derivative():
+    a = AdjFloat(1.0)
+    b = AdjFloat(2.0)
+
+    J = a*b
+    c = Control(a)
+    assert isinstance(compute_derivative(J, c), AdjFloat)
+    assert isinstance(compute_derivative(J, [c]), list)
```
connorjward: Should there be a `compute_tlm` test too?

JHopeCollins: No, these tests are checking that list-ness will be maintained, even for len==1. The result of the TLM is never a list, just a single functional, so there's nothing to check.

PR #241, opened 2026-02-04
Implementation of a Parametrised Reduced Functional

divijghose:
## Current method to use parameters
The `derivative_components` optional argument of `ReducedFunctional` is used after adding parameters to the list of controls, to specify which components are to be zeroed out (by omitting them from `derivative_components`). This allows the user to update parameters by calling the Reduced Functional while zeroing out the gradient with respect to the parameters. 

## Parametrised Reduced Functional
`ParametrisedReducedFunctional` is a subclass of `ReducedFunctional` with wrapping `call` and `derivative` methods, with the parameters as attributes. The `parameter_update` method is called to update parameters. The parameters are not included in the derivative calculation and the optional argument `derivative_components` is not required.

### DIFF
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -440,16 +440,10 @@ class ParametrisedReducedFunctional(AbstractReducedFunctional):
         self._controls = Enlist(controls)
         self._parameters = Enlist(parameters)
         self.n_opt = len(self._controls)
-        self.derivative_components = tuple(range(self.n_opt)) # Tuple of indices corresponding to optimization controls which are included in derivative calculations.
-
-
         self._all_controls= self._controls + Enlist(Control(parameters))
-
-       
-
+        
         self._reduced_functional = ReducedFunctional(functional=functional, 
                                                      controls=self._all_controls, 
-                                                     derivative_components=self.derivative_components, 
                                                      scale=scale, 
                                                      tape=tape,
                                                      eval_cb_pre=eval_cb_pre, 
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -462,7 +462,7 @@ class ParametrisedReducedFunctional(AbstractReducedFunctional):
         return self._controls
     
     @property
-    def parameters(self) -> list[Control]:
+    def parameters(self) -> list[OverloadedType]:
         return self._parameters
 
     @no_annotations
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -380,7 +380,7 @@ class ReducedFunctional(AbstractReducedFunctional):
             for control in self.controls:
                 control.unmark_as_control()
 
-class ParametrisedReducedFunctional(ReducedFunctional):
+class ParametrisedReducedFunctional(AbstractReducedFunctional):
     """Class representing the reduced functional with parameters.
 
     A reduced functional maps a control value to the provided functional.
@@ -422,7 +422,6 @@ class ParametrisedReducedFunctional(ReducedFunctional):
         tlm_cb_post (function): Callback function after evaluating the tangent linear model.
             Inputs are the functional, the tlm result, and controls.
     """
-
     def __init__(self, functional, controls, parameters,
                  scale=1.0, tape=None,
                  eval_cb_pre=lambda *args: None,
@@ -434,63 +433,190 @@ class ParametrisedReducedFunctional(ReducedFunctional):
                  hessian_cb_post=lambda *args: None,
                  tlm_cb_pre=lambda *args: None,
                  tlm_cb_post=lambda *args: None):
+        if not isinstance(functional, OverloadedType):
+            raise TypeError("Functional must be an OverloadedType.")
+        if parameters is None:
+            raise ValueError("Parameters must be provided. If no parameters are needed, use ReducedFunctional instead.")
+        print("using the new code now")
 
-
-
-
+        self.functional = functional
+        self.tape = get_working_tape() if tape is None else tape
+        self._controls = Enlist(controls)
         self._parameters = Enlist(parameters)
-        controls = Enlist(controls)
-        self.n_opt = len(controls) 
-        derivative_components = tuple(range(self.n_opt)) # Tuple of indices corresponding to optimization controls which are included in derivative calculations.
+        self.n_opt = len(self._controls)
+        self.derivative_components = tuple(range(self.n_opt)) # Tuple of indices corresponding to optimization controls which are included in derivative calculations.
+        self.scale = scale
+        self.eval_cb_pre = eval_cb_pre
+        self.eval_cb_post = eval_cb_post
+        self.derivative_cb_pre = derivative_cb_pre
+        self.derivative_cb_post = derivative_cb_post
+        self.hessian_cb_pre = hessian_cb_pre
+        self.hessian_cb_post = hessian_cb_post
+        self.tlm_cb_pre = tlm_cb_pre
+        self.tlm_cb_post = tlm_cb_post
 
-        # Prepare controls + parameters list for base class. By default, parameters are appended after the optimization controls.
-        all_controls = controls + Enlist(Control(parameters))
+        self._all_controls= Enlist(self._controls + Enlist(Control(parameters))) 
 
-        super().__init__(functional=functional,
-                         controls=all_controls,
-                         derivative_components=derivative_components,
-                         scale=scale,
-                         tape=tape,
-                         eval_cb_pre=eval_cb_pre,
-                         eval_cb_post=eval_cb_post,
-                         derivative_cb_pre=derivative_cb_pre,
-                         derivative_cb_post=derivative_cb_post,
-                         hessian_cb_pre=hessian_cb_pre,
-                         hessian_cb_post=hessian_cb_post,
-                         tlm_cb_pre=tlm_cb_pre,
-                         tlm_cb_post=tlm_cb_post)
+        self.derivative_cb_pre = _get_extract_derivative_components(self.derivative_components)
+        self.derivative_cb_post = _get_pack_derivative_components(self._all_controls, self.derivative_components)
 
 
-
+    @property
+    def controls(self) -> list[Control]:
+        return self._controls
     
     @property
     def parameters(self) -> list[Control]:
         return self._parameters
     
+    @property
+    def all_controls(self) -> list[Control]:
+        return self._all_controls
+    
+    @no_annotations
+    def derivative(self, adj_input=1.0, apply_riesz=False):
+        values = [c.tape_value() for c in self.all_controls]
+        controls = self.derivative_cb_pre(self.all_controls)
+
+        if not controls:
+            raise ValueError("""Note that the callback interface
+            for derivative_cb_pre has changed. It should now return a
+            list of controls (usually the same list as input.""")
+        
+        # Make sure `adj_input` is an OverloadedType
+        adj_input = create_overloaded_object(adj_input)
+        adj_value = adj_input._ad_mul(self.scale)
+
+        derivatives = compute_derivative(self.functional, 
+                                         controls, 
+                                         tape=self.tape,
+                                         adj_value=adj_value,
+                                         apply_riesz=apply_riesz)
+        
+        # Call callback
+        derivatives = self.derivative_cb_post(
+            self.functional.block_variable.checkpoint,
+            derivatives,
+            values)
+        
+        if not derivatives:
+            raise ValueError("""Note that the callback interface
+            for derivative_cb_post has changed. It should now return a
+            list of derivatives, usually the same list as input.""")
+        
+        return Enlist(self.all_controls.delist(derivatives))[:self.n_opt] 
+    
     @no_annotations
     def parameter_update(self, new_parameters):
         if len(Enlist(new_parameters)) != len(self._parameters):
             raise ValueError(
-                "new_parameters should be a list of same length as parameters."
+                """new_parameters should be a list of the same 
+                length as parameters."""
             )
         self._parameters = Enlist(new_parameters)
 
 
     @no_annotations
-    def derivative(self, adj_input=1.0, apply_riesz=False):
-        derivatives_full = super().derivative(adj_input=adj_input, apply_riesz=apply_riesz)
-        # Return only derivatives corresponding to optimization controls
-        return Enlist(derivatives_full)[:self.n_opt]
+    def hessian(self, m_dot, hessian_input=None, evaluate_tlm=True, apply_riesz=False):
+        # Call callback
+        values = [c.tape_value() for c in self.all_controls]
+        self.hessian_cb_pre(self.all_controls.delist(values))
 
+        r = compute_hessian(self.functional, self.all_controls, m_dot,
+                            hessian_input=hessian_input, tape=self.tape,
+                            evaluate_tlm=evaluate_tlm, apply_riesz=apply_riesz)
+
+        # Call callback
+        self.hessian_cb_post(self.functional.block_variable.checkpoint,
+                             self.all_controls.delist(r),
+                             self.all_controls.delist(values))
+
+        return self.all_controls.delist(r)
+    
 
     @no_annotations
+    def tlm(self, m_dot):
+        # Call callback
+        values = [c.tape_value() for c in self.all_controls]
+        self.tlm_cb_pre(self.all_controls.delist(values))
+
+        tlm = compute_tlm(self.functional, self.all_controls, m_dot, tape=self.tape)
+
+        # Call callback
+        self.tlm_cb_post(self.functional.block_variable.checkpoint,
+                         tlm, self.all_controls.delist(values))
+        return tlm
+    
+    @no_annotations
     def __call__(self, values):
+
         values = Enlist(values)
-        if len(values) != self.n_opt:
+        if len(values) != len(self.controls):
             raise ValueError("Length of values passed to ParametrisedReducedFunctional" \
             " must match the number of optimization controls.")
-        # concatenate optimization controls + parameters
-        full_values = values + self._parameters
-        return super().__call__(full_values)
+
+        all_values = values + self._parameters 
+ 
+
+        for i, value in enumerate(all_values):
+            control_type = type(self.all_controls[i].control) 
+            if isinstance(value, (int, float)) and control_type is AdjFloat:
+                value = self.all_controls[i].control._ad_init_object(value)
+            elif not isinstance(value, control_type):
+                if len(all_values) == 1:
+                    raise TypeError(
+                        "Control value must be an `OverloadedType` object with the same "
+                        f"type as the control, which is {control_type}"
+                    )
+                else:
+                    raise TypeError(
+                        f"The control at index {i} must be an `OverloadedType` object "
+                        f"with the same type as the control, which is {control_type}"
+                    )
+                # concatenate optimization controls + parameters
+        # Call callback.
+        self.eval_cb_pre(self.all_controls.delist(all_values))
 
 
+        for i, value in enumerate(all_values):
+            self.all_controls[i].update(value)
+
+        self.tape.reset_blocks()
+        blocks = self.tape.get_blocks()
+        self.tape._recompute_count += 1
+        with self.marked_controls():
+            with stop_annotating():
+                if self.tape._checkpoint_manager:
+                    self.tape._checkpoint_manager.recompute(self.functional)
+                else:
+                    for i in self.tape._bar("Evaluating functional").iter(
+                        range(len(blocks))
+                    ):
+                        blocks[i].recompute()
+
+        # ReducedFunctional can result in a scalar or an assembled 1-form
+        func_value = self.functional.block_variable.saved_output
+        # Scale the underlying functional value
+        func_value *= self.scale
+
+        # Call callback
+        self.eval_cb_post(func_value, self.all_controls.delist(all_values))
+        return func_value
+    
+    def optimize_tape(self):
+        self.tape.optimize(
+            controls=self.all_controls,
+            functionals=[self.functional]
+        )
+
+    @contextmanager
+    def marked_controls(self):
+        """Return a context manager which marks the active controls."""
+        for control in self.all_controls:
+            control.mark_as_control()
+        try:
+            yield
+        finally:
+            for control in self.all_controls:
+                control.unmark_as_control()
+    
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -466,7 +466,7 @@ class ParametrisedReducedFunctional(AbstractReducedFunctional):
         return self._parameters
 
     @no_annotations
-    def parameter_update(self, new_parameters):
+    def update_parameters(self, new_parameters):
         if len(Enlist(new_parameters)) != len(self._parameters):
             raise ValueError(
                 """new_parameters should be a list of the same 
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -489,7 +489,7 @@ class ParametrisedReducedFunctional(AbstractReducedFunctional):
             " must match the number of optimization controls.")
         # concatenate optimization controls + parameters
         full_values = values + self._parameters
-        return self._reduced_functional.__call__(full_values)
+        return self._reduced_functional(full_values)
     
 
 
--- a/pyadjoint/reduced_functional.py
+++ b/pyadjoint/reduced_functional.py
@@ -379,3 +379,118 @@ class ReducedFunctional(AbstractReducedFunctional):
         finally:
             for control in self.controls:
                 control.unmark_as_control()
+
+class ParametrisedReducedFunctional(ReducedFunctional):
+    """Class representing the reduced functional with parameters.
+
+    A reduced functional maps a control value to the provided functional.
+    It may also be used to compute the derivative of the functional with
+    respect to the control. In addition, parameters may be specified which
+    are updated, but not included in the derivative calculations.
+
+    Args:
+        functional (:obj:`OverloadedType`): An instance of an OverloadedType,
+            usually :class:`AdjFloat`. This should be the return value of the
+            functional you want to reduce.
+        controls (list[Control]): A list of Control instances, which you want
+            to map to the functional. It is also possible to supply a single
+            Control instance instead of a list.
+        parameters (list): A list of parameters, which are updated, but not included in the derivative.
+        scale (float): A scaling factor applied to the functional and its
+            gradient with respect to the control.
+        tape (Tape): A tape object that the reduced functional will use to
+            evaluate the functional and its gradients (or derivatives).
+        eval_cb_pre (function): Callback function before evaluating the
+            functional. Input is a list of Controls.
+        eval_cb_pos (function): Callback function after evaluating the
+            functional. Inputs are the functional value and a list of Controls.
+        derivative_cb_pre (function): Callback function before evaluating
+            derivatives. Input is a list of Controls.
+            Should return a list of Controls (usually the same
+            list as the input) to be passed to compute_derivative.
+        derivative_cb_post (function): Callback function after evaluating
+            derivatives.  Inputs are: functional.block_variable.checkpoint,
+            list of functional derivatives, list of functional values.
+            Should return a list of derivatives (usually the same
+            list as the input) to be returned from self.derivative.
+        hessian_cb_pre (function): Callback function before evaluating the Hessian.
+            Input is a list of Controls.
+        hessian_cb_post (function): Callback function after evaluating the Hessian.
+            Inputs are the functional, a list of Hessian, and controls.
+        tlm_cb_pre (function): Callback function before evaluating the tangent linear model.
+            Input is a list of Controls.
+        tlm_cb_post (function): Callback function after evaluating the tangent linear model.
+            Inputs are the functional, the tlm result, and controls.
+    """
+
+    def __init__(self, functional, controls, parameters,
+                 scale=1.0, tape=None,
+                 eval_cb_pre=lambda *args: None,
+                 eval_cb_post=lambda *args: None,
+                 derivative_cb_pre=lambda controls: controls,
+                 derivative_cb_post=lambda checkpoint, derivative_components,
+                 controls: derivative_components,
+                 hessian_cb_pre=lambda *args: None,
+                 hessian_cb_post=lambda *args: None,
+                 tlm_cb_pre=lambda *args: None,
+                 tlm_cb_post=lambda *args: None):
+
+
+
+
+        self._parameters = Enlist(parameters)
+        controls = Enlist(controls)
+        self.n_opt = len(controls) 
+        derivative_components = tuple(range(self.n_opt)) # Tuple of indices corresponding to optimization controls which are included in derivative calculations.
+
+        # Prepare controls + parameters list for base class. By default, parameters are appended after the optimization controls.
+        all_controls = controls + Enlist(Control(parameters))
+
+        super().__init__(functional=functional,
+                         controls=all_controls,
+                         derivative_components=derivative_components,
+                         scale=scale,
+                         tape=tape,
+                         eval_cb_pre=eval_cb_pre,
+                         eval_cb_post=eval_cb_post,
+                         derivative_cb_pre=derivative_cb_pre,
+                         derivative_cb_post=derivative_cb_post,
+                         hessian_cb_pre=hessian_cb_pre,
+                         hessian_cb_post=hessian_cb_post,
+                         tlm_cb_pre=tlm_cb_pre,
+                         tlm_cb_post=tlm_cb_post)
+
+
+
+    
+    @property
+    def parameters(self) -> list[Control]:
+        return self._parameters
+    
+    @no_annotations
+    def parameter_update(self, new_parameters):
+        if len(Enlist(new_parameters)) != len(self._parameters):
+            raise ValueError(
+                "new_parameters should be a list of same length as parameters."
+            )
+        self._parameters = Enlist(new_parameters)
+
+
+    @no_annotations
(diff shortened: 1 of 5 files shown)


divijghose:
In this intermediate implementation, the `ParametrisedReducedFunctional` inherits from `AbstractReducedFunctional`. This means that `self.controls` returns `user_controls`, which is the required behaviour as discussed above.

However, a lot of the code is duplicated, and, as @colinjcotter suggested, a more efficient way to do this would be to call `ReducedFunctional` inside `ParametrisedReducedFunctional`, which will get the required behaviour from `self.controls` while reusing most of the methods from `ReducedFunctional`.

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+            Should return a list of derivatives (usually the same
+            list as the input) to be returned from self.derivative.
+        hessian_cb_pre (function): Callback function before evaluating the Hessian.
+            Input is a list of Controls.
+        hessian_cb_post (function): Callback function after evaluating the Hessian.
+            Inputs are the functional, a list of Hessian, and controls.
+        tlm_cb_pre (function): Callback function before evaluating the tangent linear model.
+            Input is a list of Controls.
+        tlm_cb_post (function): Callback function after evaluating the tangent linear model.
+            Inputs are the functional, the tlm result, and controls.
```
JHopeCollins: We should have a think about whether these should be for `all_controls` (as it is currently), or just the `ParameterisedReducedFunctional.controls`, and possibly add separate callbacks for the parameters.

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+    """
+    def __init__(self, functional, controls, parameters,
+                 scale=1.0, tape=None,
+                 eval_cb_pre=lambda *args: None,
+                 eval_cb_post=lambda *args: None,
+                 derivative_cb_pre=lambda controls: controls,
+                 derivative_cb_post=lambda checkpoint, derivative_components,
+                 controls: derivative_components,
```
JHopeCollins, line 449: Repeated arg in the middle of the kwargs?
```suggestion
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+                 hessian_cb_pre=lambda *args: None,
+                 hessian_cb_post=lambda *args: None,
+                 tlm_cb_pre=lambda *args: None,
+                 tlm_cb_post=lambda *args: None):
+        if not isinstance(functional, OverloadedType):
+            raise TypeError("Functional must be an OverloadedType.")
+        if parameters is None:
+            raise ValueError("Parameters must be provided. If no parameters are needed, use ReducedFunctional instead.")
```
JHopeCollins, line 463: Does passing an empty list work?

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+        self._controls = Enlist(controls)
+        self._parameters = Enlist(parameters)
+        self.n_opt = len(self._controls)
+        self._all_controls= self._controls + Enlist([Control(p) if not isinstance(p, Control) else p for p in self._parameters])
```
JHopeCollins: Do we want users passing parameters as a `Control` or is that leaking the implementation into the interface, and the parameters should be `OverloadedType`?

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+
+
+
+    @property
+    def controls(self) -> list[Control]:
+        return self._controls
+    
+    @property
+    def parameters(self) -> list[Control]:
+        return self._parameters
```
JHopeCollins:
```suggestion
    def parameters(self) -> list[OverloadedType]:
        return self._parameters
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+                length as parameters."""
+            )
+        self._parameters = Enlist(new_parameters)
+    
+    @no_annotations
+    def derivative(self, adj_input=1.0, apply_riesz=False):
+        derivatives_all = self._reduced_functional.derivative(adj_input=adj_input, 
+                                                                    apply_riesz=apply_riesz)
+        
+        return Enlist(derivatives_all)[:self.n_opt]
```
JHopeCollins: If the user didn't give us a list of controls then we shouldn't return a list of derivatives
```suggestion
        return self.controls.delist(Enlist(derivatives_all)[:self.n_opt])
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+        values = Enlist(values)
+        if len(values) != self.n_opt:
+            raise ValueError("Length of values passed to ParametrisedReducedFunctional" \
+            " must match the number of optimization controls.")
+        # concatenate optimization controls + parameters
+        full_values = values + self._parameters
+        return self._reduced_functional.__call__(full_values)
+    
+
+
```
JHopeCollins:
```suggestion
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+    
+    @no_annotations
+    def __call__(self, values):
+        values = Enlist(values)
+        if len(values) != self.n_opt:
+            raise ValueError("Length of values passed to ParametrisedReducedFunctional" \
+            " must match the number of optimization controls.")
+        # concatenate optimization controls + parameters
+        full_values = values + self._parameters
+        return self._reduced_functional.__call__(full_values)
```
JHopeCollins: You rarely need to call dunder methods explicitly
```suggestion
        return self._reduced_functional(full_values)
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+
+
+    @no_annotations
+    def hessian(self, m_dot, hessian_input=None, evaluate_tlm=True, apply_riesz=False):
+        hessian_all = self._reduced_functional.hessian(m_dot, 
+                                                              hessian_input=hessian_input, 
+                                                              evaluate_tlm=evaluate_tlm, 
+                                                              apply_riesz=apply_riesz)
+
+        return hessian_all[:self.n_opt] # Return only the hessian components corresponding to optimization controls.
```
JHopeCollins:
```suggestion
		# Return only the hessian components corresponding to optimization controls.
        return self.controls.delist(Enlist(hessian_all)[:self.n_opt])
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+    
+
+
+
+    @no_annotations
+    def hessian(self, m_dot, hessian_input=None, evaluate_tlm=True, apply_riesz=False):
+        hessian_all = self._reduced_functional.hessian(m_dot, 
+                                                              hessian_input=hessian_input, 
+                                                              evaluate_tlm=evaluate_tlm, 
+                                                              apply_riesz=apply_riesz)
```
JHopeCollins, line 527: `self._reduced_functional.hessian` will expect `len(m_dot) == len(self._all_controls)` so I think you will have to pad it with zeros (but you should write out the maths to check this will DTRT). Something like:
```py
m_dot_all = Enlist(m_dot) + [p._ad_init_zero() for p in self.parameters]
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+
+        return hessian_all[:self.n_opt] # Return only the hessian components corresponding to optimization controls.
+    
+
+    @no_annotations
+    def tlm(self, m_dot):
+        tlm_all = self._reduced_functional.tlm(m_dot)
+        return tlm_all[:self.n_opt] # Return only the tlm components corresponding to optimization controls.
```
JHopeCollins: 1. You'll need the same zero-padding of `m_dot` here as in the `hessian`.
2. The result of `tlm` lives in the same space as the `functional` so you don't need to apply the mask.

```suggestion
    @no_annotations
    def tlm(self, m_dot):
        tlm_all = self._reduced_functional.tlm(m_dot)
        return tlm
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+    
+
+    def optimize_tape(self):
+        self._reduced_functional.tape.optimize(
+            controls=self._reduced_functional.controls,
+            functionals=[self._reduced_functional.functional]
+        )
```
JHopeCollins: Surely this is equivalent?
```suggestion
    def optimize_tape(self):
        self._reduced_functional.optimize_tape()
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+
+
```
JHopeCollins:
```suggestion
```

tests/pyadjoint/test_parametrised_rf.py
```diff
@@ -0,0 +1,383 @@
...
+    # Test parameter update
+    new_p = p_val * mult_factor
+    Jhat.parameter_update(new_p)
+    result = Jhat(c_val)
+    expected = c_val * new_p
+    assert result == expected
+    
+    # Test derivative
+    deriv = Jhat.derivative()
+    assert deriv[0] == new_p
```
JHopeCollins: Please can you replace anywhere you're testing the `derivative()` value directly with a call to `taylor_to_dict` and checking the convergence rates?

tests/pyadjoint/test_parametrised_rf.py
```diff
@@ -0,0 +1,383 @@
...
+    result = Jhat(c_val)
+    expected = c_val * p_val
+    assert result == expected
+    
+    # Test parameter update
+    new_p = p_val * mult_factor
+    Jhat.parameter_update(new_p)
+    result = Jhat(c_val)
+    expected = c_val * new_p
+    assert result == expected
```
JHopeCollins, line 73: This is a good start, but it'd be good to test a couple more cases. For example:
1. Update the parameters twice before calling `Jhat(m)`. Is the most recent one used?
2. Call `Jhat(m)` multiple times before resetting the parameters.

For the derivative, it'd be good to test that the derivative uses the values of the parameters from the last time `Jhat(m)` was called, which I think you could do like so (this should be the only time you'll need to test `derivative` directly rather than using the taylor test:

1. Set the parameters
2. Call `Jhat(m)`
3. Check the value of `Jhat.derivative()` is correct.
4. Call `Jhat.update_parameters(params)`
5. Check that the result of calling `Jhat.derivative()` hasn't changed.

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
+    @property
+    def controls(self) -> list[Control]:
+        return self._controls
+    
+    @property
+    def parameters(self) -> list[Control]:
+        return self._parameters
+
+    @no_annotations
+    def parameter_update(self, new_parameters):
```
JHopeCollins: Methods that do an action usually have imperative names
```suggestion
    def update_parameters(self, new_parameters):
```

tests/pyadjoint/test_parametrised_rf.py
```diff
@@ -0,0 +1,383 @@
...
+    
+    # Test initial evaluation
+    result = Jhat([c1_new, c2_new])
+    expected = (c1_new + c2_new) ** 2 * p1_val - c1_new * p2_val
+    assert result == expected
+    
+    # Update and test again
+    Jhat.parameter_update([p1_new, p2_new])
+    result = Jhat([c1_new, c2_new])
+    expected = (c1_new + c2_new) ** 2 * p1_new - c1_new * p2_new
```
JHopeCollins: It would be more robust if the `build_*` functions didn't create the `ParameterisedReducedFunctional` but instead just did the expression, then this line (and other similar ones) could just be something like:

```py
Jhat.update_parameters(new_params)
assert Jhat(m) == complex_expression(m+new_params)
```

tests/pyadjoint/test_parametrised_rf.py
```diff
@@ -0,0 +1,383 @@
...
+    (4.0, 2.5, 0.5),
+])
+def test_parametrised_rf_basic(c_val, p_val, mult_factor):
+    """Test basic evaluation of parametrised reduced functional with various values."""
+    Jhat= build_single_control_single_param(c_val, p_val)
+    
+    # Test initial evaluation
+    result = Jhat(c_val)
+    expected = c_val * p_val
+    assert result == expected
```
JHopeCollins: Comparing equality of floats is unreliable at best, you should use `np.isclose` with a tight tolerance instead.

JHopeCollins (changes requested):
This is shaping up nicely!

I've left a few comments, but my main two requests are for a bit more test coverage.

1. Please can you test the convergence rates from `taylor_to_dict`? This will do both the first and second order taylor tests so that we can check that the tlm and hessian are correct.
2. Please can you add an optimisation test? You could just use a quadratic polynomial as the objective functional with the coefficients being parameters. Then the optimisation should be globally convergent and you can test the result against the analytic minimum.
If you want to be very thorough then you could also test a multivariate quadratic to make sure that it works with multiple controls.

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,141 @@ def marked_controls(self):
...
             for control in self.controls:
                 control.unmark_as_control()
+
+class ParametrisedReducedFunctional(AbstractReducedFunctional):
+    """Class representing the reduced functional with parameters.
+
+    A reduced functional maps a control value to the provided functional.
+    It may also be used to compute the derivative of the functional with
+    respect to the control. In addition, parameters may be specified which
+    are updated, but not included in the derivative calculations.
```
JHopeCollins: It'd be nice to show a few lines of maths to explain here like in the docstring of `AbstractReducedFunctional`

pyadjoint/reduced_functional.py
divijghose: Thanks! I've changed the method name and updated it in the tests as well.

pyadjoint/reduced_functional.py
divijghose: Yes, they're equivalent and `optimize_tape` has been updated accordingly. Thanks!

tests/pyadjoint/test_parametrised_rf.py
divijghose: `test_parametrised_rf_multiple_update_parameters` covers the first example that you have mentioned. Thanks for the suggestions of the other tests, I will add them and resolve this conversation.

pyadjoint/reduced_functional.py
divijghose: @JHopeCollins, the argument list is the similar to the one used in `ReducedFunctional`, with `derivative_components` omitted and `parameters` added.

pyadjoint/reduced_functional.py
divijghose: Passing an empty list should not work. I have updated `ParametrisedReducedFunctional` to check for it:
```python3
        if len(Enlist(parameters)) == 0:
            raise ValueError("Parameters list cannot be empty. If no parameters are needed, use ReducedFunctional instead.")
```

along with a corresponding test
```python3
def test_parametrised_rf_empty_parameter_list():
    """Test that creating a ParametrisedReducedFunctional with an empty parameter list raises an error."""
    c = AdjFloat(2.0)
    J = c * 3.0 
    with pytest.raises(ValueError):
        Jhat = ParametrisedReducedFunctional(J, Control(c), parameters=[])
```

pyadjoint/reduced_functional.py
divijghose: Users should not pass parameters as `Control`, that line was redundant and has been removed.

pyadjoint/reduced_functional_numpy.py
```diff
@@ -129,6 +130,132 @@ def set_controls(self, array):
             control.update(m_i)
         return m
 
+class ParametrisedReducedFunctionalNumPy(AbstractReducedFunctional):
```
JHopeCollins: Why does `prf_np = ReducedFunctionalNumPy(parameterised_rf)` not work? We should fix that instead of duplicating all this code. In theory the numpy rf should work for any `AbstractReducedFunctional`.

divijghose: I've pushed a change to remove `ParametrisedReducedFunctionalNumPy` and instead just use `ReducedFunctionalNumPy(parametrised_rf)` in the optimisation. Thanks for noticing this!

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,164 @@ def marked_controls(self):
         finally:
             for control in self.controls:
                 control.unmark_as_control()
+
+class ParametrisedReducedFunctional(AbstractReducedFunctional):
+    """Class representing the reduced functional with parameters.
+    An object which encompasses computations of the form::
+
+        Jhat(m, p) = J(u(m, p), m, p)
```
JHopeCollins: These are parameters not arguments
```suggestion
        Jhat(m; p) = J(u(m; p), m; p)
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,164 @@ def marked_controls(self):
...
+
+    Where `u` is the system state and `m` is a `pyadjoint.Control` or list of
+    `pyadjoint.Control`, `p` is a list of parameters, `J` is an overloaded type providing 
+    the functional value, and `Jhat` is a reduced functional where the explicit dependence 
+    on `u` has been eliminated. The parameters `p` can be updated, after which evaluation 
+    of the reduced functional is performed with the updated parameters, but they are not 
+    included in the derivative calculations.
+
+    A parametrised reduced functional is callable and maps a control value and parameter/s
```
JHopeCollins:
```suggestion
    A parametrised reduced functional is callable and maps a control value and parameters 
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,164 @@ def marked_controls(self):
...
+class ParametrisedReducedFunctional(AbstractReducedFunctional):
+    """Class representing the reduced functional with parameters.
+    An object which encompasses computations of the form::
+
+        Jhat(m, p) = J(u(m, p), m, p)
+
+    Where `u` is the system state and `m` is a `pyadjoint.Control` or list of
+    `pyadjoint.Control`, `p` is a list of parameters, `J` is an overloaded type providing 
+    the functional value, and `Jhat` is a reduced functional where the explicit dependence 
+    on `u` has been eliminated. The parameters `p` can be updated, after which evaluation
```
JHopeCollins:
```suggestion
    on `u` has been eliminated. The parameters `p` can be updated between evaluations, after which evaluation 
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,164 @@ def marked_controls(self):
...
+    Where `u` is the system state and `m` is a `pyadjoint.Control` or list of
+    `pyadjoint.Control`, `p` is a list of parameters, `J` is an overloaded type providing 
+    the functional value, and `Jhat` is a reduced functional where the explicit dependence 
+    on `u` has been eliminated. The parameters `p` can be updated, after which evaluation 
+    of the reduced functional is performed with the updated parameters, but they are not 
+    included in the derivative calculations.
+
+    A parametrised reduced functional is callable and maps a control value and parameter/s 
+    to the provided functional. It may also be used to compute the derivative of the functional 
+    with respect to the control. In addition, parameters can be updated, but are not included
```
JHopeCollins:
```suggestion
    with respect to the control. In addition, parameters can be updated between evaluations, but are not included 
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,164 @@ def marked_controls(self):
...
+            to map to the functional. It is also possible to supply a single
+            Control instance instead of a list.
+        parameters (list): A list of parameters, which are updated, but not included in the derivative.
+        scale (float): A scaling factor applied to the functional and its
+            gradient with respect to the control.
+        tape (Tape): A tape object that the reduced functional will use to
+            evaluate the functional and its gradients (or derivatives).
+        eval_cb_pre (function): Callback function before evaluating the
+            functional. Input is a list of Controls.
+        eval_cb_pos (function): Callback function after evaluating the
```
JHopeCollins:
```suggestion
        eval_cb_post (function): Callback function after evaluating the
```

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,164 @@ def marked_controls(self):
...
+            functional. Inputs are the functional value and a list of Controls.
+        derivative_cb_pre_for_controls (function): Callback function before evaluating
```
JHopeCollins: This is a very large number of arguments for init! Why not keep `derivative_cb_pre` etc and make the calling signature `derivative_cb_pre(controls, parameters)` instead of jut `derivative_cb_pre(controls)` like in `ReducedFunctional`?

pyadjoint/reduced_functional.py
```diff
@@ -379,3 +379,164 @@ def marked_controls(self):
...
+                 hessian_cb_pre_for_parameters=lambda *args: None,
+                 hessian_cb_post_for_parameters=lambda *args: None,
+                 tlm_cb_pre_for_parameters=lambda *args: None,
+                 tlm_cb_post_for_parameters=lambda *args: None):
+        if not isinstance(functional, OverloadedType):
+            raise TypeError("Functional must be an OverloadedType.")
+        if parameters is None:
+            raise ValueError("Parameters must be provided. If no parameters are needed, use ReducedFunctional instead.")
+        if len(Enlist(parameters)) == 0:
+            raise ValueError("Parameters list cannot be empty. If no parameters are needed, use ReducedFunctional instead.")
```
JHopeCollins: Personally I would be fine with allowing zero parameters, but this is an API choice and could be something to discuss in the meeting.

tests/pyadjoint/test_parametrised_rf.py
```diff
@@ -0,0 +1,486 @@
...
+def multi_control_single_param_expr(c1_val, c2_val, p_val):
+    return c1_val**3 * c2_val**4 * p_val
+
+def single_control_multi_param_expr(c_val, p1_val, p2_val):
+    return c_val**3 * p1_val * p2_val
+
+def multi_control_multi_param_expr(c1_val, c2_val, p1_val, p2_val):
+    return c1_val**3 * c2_val**4 * p1_val + c1_val**2 * c2_val**5 * p2_val
+
+def complex_expression(c1_val, c2_val, p1_val, p2_val):
```
JHopeCollins: "complex" is an overloaded term
```suggestion
def complicated_expression(c1_val, c2_val, p1_val, p2_val):
```

tests/pyadjoint/test_parametrised_rf.py
```diff
@@ -0,0 +1,486 @@
...
+    return (c1_val + c2_val)**3 * p1_val - c1_val**2 * c2_val**2 * p2_val 
+
+def check_taylor_test_convergence(Jhat, controls):
+    """Helper function to check that the taylor test convergence rates are as expected."""
+    h = [AdjFloat(1.0) for _ in controls]
+    taylor_results = taylor_to_dict(Jhat, controls, h)
+    assert min(taylor_results["R0"]["Rate"]) >= 0.95, f"Error in R0 rate: {taylor_results['R0']['Rate']}"
+    assert min(taylor_results["R1"]["Rate"]) >= 1.95, f"Error in R1 rate: {taylor_results['R1']['Rate']}"
+    assert min(taylor_results["R2"]["Rate"]) >= 2.95, f"Error in R2 rate: {taylor_results['R2']['Rate']}"
```
JHopeCollins: If something failed I want to see more information.
```suggestion
    assert min(taylor_results["R0"]["Rate"]) >= 0.95, f"Error in R0 rate: {taylor_results['R0']}"
    assert min(taylor_results["R1"]["Rate"]) >= 1.95, f"Error in R1 rate: {taylor_results['R1']}"
    assert min(taylor_results["R2"]["Rate"]) >= 2.95, f"Error in R2 rate: {taylor_results['R2']}"
```

tests/pyadjoint/test_parametrised_rf.py
```diff
@@ -0,0 +1,486 @@
...
+    with set_working_tape() as tape_1:
+        c = AdjFloat(c_val)
+        p = AdjFloat(p_val)
+        J = c * p
+        Jhat_rf = ReducedFunctional(J, [Control(c), Control(p)], derivative_components=[0])
+    
+    # Build parametrised reduced functional
+    with set_working_tape() as tape_2:
+        J = c * p
+        Jhat_param_rf = ParametrisedReducedFunctional(J, Control(c), parameters=p)
```
JHopeCollins: Controls and parameters should not appear on multiple tapes.
```suggestion
        c = AdjFloat(c_val)
        p = AdjFloat(p_val)
        J = c * p
        Jhat_param_rf = ParametrisedReducedFunctional(J, Control(c), parameters=p)
```

tests/pyadjoint/test_parametrised_rf.py
```diff
@@ -0,0 +1,486 @@
...
+    result_prf_new = minimize(Jhat_prf)
+    # Check that the optimal control value is close to the expected minimum of the quadratic
+    assert np.isclose(result_prf_new, new_optima, atol=1e-6)
+
+@pytest.mark.parametrize("c_val, p_val1, p_val2, p_val3, p_val1_new, p_val2_new, p_val3_new", [
+    (4.0, 3.0, 6.9, 7.4, 8.5, -3.8, 9.0),
+    (5.5, 3.4, -4.0, 15.0, 9.2, 8.4, 6.7),
+    (9.0, 2.5, 6.3, 1.0, 5.9, 0.0, -1.0),
+])
+def test_optimisation_on_quadratic_polynomial_w_TAO(c_val, p_val1, p_val2, p_val3, p_val1_new, p_val2_new, p_val3_new):
```
JHopeCollins, line 466: petsc4py is an optional dependency of pyadjoint so we need to conditionally skip this test if it isn't installed. Something like this should be fine:
```python
try:
	import petsc4py
except ImportError:
	pytest.skip(reason="petsc4py not installed")
```

divijghose:
Considering the discussion at the [Firedrake meeting](https://github.com/firedrakeproject/firedrake/wiki/Firedrake-meeting-2026-04-28), the following commits rewrite `ReducedFunctional` to accept a `parameters` argument, with the eventual aim of deprecating the use of `derivative_components`. For now, a user can pass either `parameters` or `derivative_components` (but not both together, this would throw up an error).

divijghose:
@JHopeCollins, with respect to the recent CI failure:
```shell
    @no_annotations
    def derivative(self, adj_input=1.0, apply_riesz=False):
        values = [c.tape_value() for c in self.controls]
>       controls = self.derivative_cb_pre(self.parameters if hasattr(self, "_parameters") else None, self.controls)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       TypeError: EnsembleReducedFunctional.<lambda>() takes 1 positional argument but 2 were given
```
This happens because the derivative callback now has the signature `derivative_cb_pre(parameters, controls)`, which does not reflect downstream in `EnsembleReducedFunctional`.

JHopeCollins:
> This happens because the derivative callback now has the signature derivative_cb_pre(parameters, controls), which does not reflect downstream in EnsembleReducedFunctional.

If the user has not passed parameters then we shouldn't change the callback interface otherwise we will break current code before the deprecation cycle is done. For now the callback signature will depend on whether parameters are passed or not.
I also think it makes more sense to put the controls first in the signature: `derivative_cb_pre(controls, parameters)`

divijghose:
> > This happens because the derivative callback now has the signature derivative_cb_pre(parameters, controls), which does not reflect downstream in EnsembleReducedFunctional.
> 
> If the user has not passed parameters then we shouldn't change the callback interface otherwise we will break current code before the deprecation cycle is done. For now the callback signature will depend on whether parameters are passed or not. I also think it makes more sense to put the controls first in the signature: `derivative_cb_pre(controls, parameters)`

@JHopeCollins, I've added the functions `_call_derivative_cb_pre` and `_call_derivative_db_post` to handle the callbacks wit parameters while maintaining backwards compatibility and emitting a deprecation warning. `derivative` has been updated accordingly.

pyadjoint/reduced_functional.py
```diff
@@ -306,18 +448,26 @@ def tlm(self, m_dot):
...
     def __call__(self, values):
         values = Enlist(values)
-        if len(values) != len(self.controls):
+        if hasattr(self, "_parameters") and len(values) != self.n_opt:
             raise ValueError(
-                "values should be a list of same length as controls."
+                f"values should be a list of same length as optimization controls, which is {self.n_opt}."
             )
+        elif len(values) != len(self.controls):
+            raise ValueError("values should be a list of same length as controls.")
```
JHopeCollins: Don't need to check `parameters`, just make everyone's life better with a informative error message

pyadjoint/reduced_functional.py
```diff
@@ -361,13 +511,14 @@ def __call__(self, values):
         # Call callback
         self.eval_cb_post(func_value, self.controls.delist(values))
 
-        return func_value
+        if not hasattr(self, "_parameters"):
+            return func_value
+        else:
+            full_values = values + self._parameters
+            return self._reduced_functional(full_values)
```
JHopeCollins: This case will cause the tape to be evaluated twice. Put this check at the beginning of the method so we recurse early before any recomputations.

pyadjoint/reduced_functional.py
```diff
@@ -306,18 +448,26 @@ def tlm(self, m_dot):
...
+        self.tlm_cb_post(
+            self.functional.block_variable.checkpoint, tlm, self.controls.delist(values)
+        )
+        if not hasattr(self, "_parameters"):
+            return tlm
+        else:
+            # self._reduced_functional.tlm will expect len(m_dot) = len(self._all_controls), so we pad it with zeros.
+            m_dot_all = Enlist(m_dot) + [p._ad_init_zero() for p in self._parameters]
+            tlm_all = self._reduced_functional.tlm(m_dot_all)
+            return tlm_all
```
JHopeCollins: Same as `__call__`, move this check to the beginning of the method.

pyadjoint/reduced_functional.py
```diff
@@ -156,18 +158,64 @@ def pack_derivative_components(checkpoint, derivatives, values):
...
+    if parameters is None:
+        return cb(controls)
+    try:
+        return cb(controls, parameters)
+    except TypeError:
+        warnings.warn(
+            message="derivative_cb_pre should accept (controls, parameters)."
+            "Falling back to deprecated signature (controls). ",
+            category=DeprecationWarning
+        )
```
JHopeCollins, line 181:
```suggestion
    except TypeError as err:
        raise TypeError(
            "derivative_cb_pre should accept (controls, parameters)."
        ) from err
```

pyadjoint/reduced_functional.py
```diff
@@ -261,41 +371,73 @@ def derivative(self, adj_input=1.0, apply_riesz=False):
...
 
         # Call callback
-        derivatives = self.derivative_cb_post(
+        derivatives = _call_derivative_cb_post(
+            self.derivative_cb_post,
             self.functional.block_variable.checkpoint,
             derivatives,
-            values)
+            values,
+            getattr(self, "_parameters", None),
```
JHopeCollins: Change callback signature depending on parameters like the others.
(and raise informative type error).

pyadjoint/reduced_functional.py
```diff
@@ -235,22 +300,67 @@ def __init__(self, functional, controls,
...
+                scale=scale,
+                tape=tape,
+                eval_cb_pre=eval_cb_pre,
+                eval_cb_post=eval_cb_post,
+                derivative_cb_pre=derivative_cb_pre,
+                derivative_cb_post=derivative_cb_post,
+                hessian_cb_pre=hessian_cb_pre,
+                hessian_cb_post=hessian_cb_post,
+                tlm_cb_pre=tlm_cb_pre,
+                tlm_cb_post=tlm_cb_post,
```
JHopeCollins: Callbacks for the inner RF should all be `None` because the inner one doesn't know about controls vs parameters, but the callback signature does.

JHopeCollins (changes requested):
Generally good, just the points discussed:

- Don't do unnecessary work to throw away the result.
- Minimal scope for dropping into the inner RF (only for the computation). Keep the controls vs parameters information to do the callbacks etc.
- 0 parameters shouldn't be so special. Just have an empty list rather than a non-existent attribute.
- Remove any changes that are purely formatting so it's clearer what the actual changes are.

pyadjoint/reduced_functional.py
```diff
@@ -156,18 +158,120 @@ def pack_derivative_components(checkpoint, derivatives, values):
...
-                zero_derivative *= 0.
+                zero_derivative *= 0.0
                 derivatives_out.append(zero_derivative)
         return derivatives_out
+
     return pack_derivative_components
 
 
+def _call_derivative_cb_pre(cb, controls, parameters=None):
+    """Call `derivative_cb_pre` with (controls, parameters) if parameters are passd, otherwise preserve backwards
```
dham:
```suggestion
    """Call `derivative_cb_pre` with (controls, parameters) if parameters are passed, otherwise preserve backwards
```

pyadjoint/reduced_functional.py
dham: change the default callbacks to take `(controls, parameters=None)`

pyadjoint/reduced_functional.py
divijghose: Now checking the length of passed `values` against `self._controls` regardless of parameters

PR #248, opened 2026-02-15
Fix SingleMemoryStorageSchedule checkpoint clearing during forward replay

### LINKED ISSUE #211: indiscriminately clearing checkpoints with `SingleMemoryStorageSchedule` corrupts  the adjoints
Second part of an issue we have encountered https://github.com/g-adopt/g-adopt/issues/237, consider [ the case of this reproducer](https://gist.github.com/sghelichkhani/3329b58404304c97b168f8bc37d3161c). When using no scheduler by:

```
python minimal_clearing_cache.py  none
```
we get:

```
using scheduler: NoneType
        J1: 12.520709541433629
        dJdm 1: 3205.301642607008
```
But when using a scheduler we have by:

```
python minimal_clearing_cache.py memory
```
we get:
```
using scheduler: SingleMemoryStorageSchedule
        J1: 12.520709541433629
        dJdm 1: 73765.11424588156
```

## Cause

`SingleMemoryStorageSchedule` explicitly deletes every checkpoint that did not appear in the previous step’s `adjoint_dependencies`:
 
https://github.com/dolfin-adjoint/pyadjoint/blob/a4f940a7899809a65d43d9fb9e6049a70755a30a/pyadjoint/checkpointing.py#L362-L364

If a variable is part of a long-range dependency (e.g. is used every third step), it can be missing from the immediately-preceding dependency set even though it is still required later. The checkpoint is therefore cleared and the reverse pass reconstructs an incorrect value, giving a wrong gradient. 

Considering the comment saying “Handle the case for SingleMemoryStorageSchedule”, I would have thought there is a reason for this, but no explanation is given. 

## Ideas (!?)

- Remove the special-case clearing and rely on the revised dependency machinery to decide what can be freed, or

 - Only clear if the variable is guaranteed not to re-appear later (e.g. by inspecting timesteps[step:].adjoint_dependencies) ?

### LINKED ISSUE #260: SingleMemoryStorageSchedule clears checkpoints at last forward use and corrupts gradients when the functional is re-evaluated at a new control
This is a follow-up to #211 and is independent of the fix in #248: the reproducer below gives
the same wrong gradient on current `master` (710d7d0) with or without #248 applied.

Consider this script:

```python
from firedrake import *
from firedrake.adjoint import *
from checkpoint_schedules import SingleMemoryStorageSchedule

continue_annotation()
tape = get_working_tape()
tape.enable_checkpointing(SingleMemoryStorageSchedule())

mesh = SquareMesh(1, 1, 1, quadrilateral=True)
V = FunctionSpace(mesh, "CG", 1)

m = Function(V).assign(1.0)
f = Function(V)
u = Function(V)

for i in tape.timestepper(iter(range(4))):
    if i == 0:
        f.project(2 * m)
        u.project(m)
    elif i == 2:
        u.project(u * f)
    else:
        u.project(Constant(1.1) * u)

J = assemble(u ** 2 * dx)
pause_annotation()

rf = ReducedFunctional(J, Control(m))

m_new = Function(V).assign(2.0)
print("J(m=2):      ", rf(m_new))
dJdm = rf.derivative().riesz_representation("L2")
print("|dJdm|(m=2): ", sqrt(assemble(dJdm ** 2 * dx)))
```

This prints

```
J(m=2):       93.70239999999976
|dJdm|(m=2):  140.55359999999965
```

The functional is right but the gradient is wrong. Everything here is simple enough to do by
hand: f = 2m and u goes m -> 1.1m -> 2.2m^2 -> 2.42m^2, so J = 5.8564 m^4 and
dJ/dm = 23.4256 m^3, which at m = 2 is 187.4048. And indeed, if you comment out the
`enable_checkpointing` line you get exactly that. If you instead evaluate at the taped control
value (`m_new.assign(1.0)`), the scheduled and unscheduled gradients agree again — the bug
only shows when the reduced functional is evaluated away from the point at which the tape
was recorded, which is of course what every optimisation loop does.

## Cause

During the forward replay triggered by `rf(m_new)`, the `SingleMemoryStorageSchedule` branch
of the clear-down loop in `CheckpointManager` deletes the checkpoint of every block variable
at its last forward use:

https://github.com/dolfin-adjoint/pyadjoint/blob/710d7d0a8b29052b5fdfa11824e22fc4cf4ee6a9/pyadjoint/checkpointing.py#L362-L365

This relies on `BlockVariable.saved_output` falling back to the live Function when the
checkpoint is `None`. The live Function, however, still holds the value from *taping*, not
from the replay. In the script, f is recomputed correctly (f = 4) into its checkpoint during
the replay, but the clear-down at step 2 — its last forward use — deletes it again. The
adjoint of `u.project(u * f)` then needs f as the linearisation point for the derivative with
respect to u, reads `saved_output`, and gets the stale taped value f = 2. The factor-of-two
error in that one weight is exactly the gap between 140.5536 and 187.4048.

Two changes to the script make the wrong answer "accidentally" right, which shows how narrow
the existing test coverage is. Remove `u.project(m)` at step 0 and the gradient is correct,
because the stale weight then multiplies a path that carries no control dependence (the
derivative with respect to f itself needs u's value, not f's, and u's checkpoint survives
since u is superseded). Move the `u.project(u * f)` from step 2 to step 1 and current
`master` is also correct, but only because the old condition's `step > 1` guard skips the
clearing there — with #248 applied that accidental protection goes away. None of our tests
evaluate at a new control under this schedule, so none of this is caught.

The underlying problem is that "last forward use" is not "last use": the adjoint of that step
may still need the variable as a linearisation point, and under `StorageType.WORK` the block
variable's own checkpoint slot is the only place the recomputed value lives. The obvious
repair — also require the variable not to be in the step's `adjoint_dependencies` before
clearing — runs into the same difficulty that caused #211 in the first place, namely that
`adjoint_dependencies` is only fully populated during the reverse pass, so on the first
evaluation after taping it would under-protect. Perhaps the clearing should only trust
`adjoint_dependencies` once they have been revised, and conservatively keep terminal
variables before that; or perhaps the silent fallback of `saved_output` to the live object is
the thing to rethink. Happy to put this on the agenda for the dev meeting.

sghelichkhani:
## Summary

Fixes #211. Fixes #260.

Both issues come from the same spot: during forward replay, the `SingleMemoryStorageSchedule` branch of the checkpoint clearing decides at a variable's last forward use whether its checkpoint can be discarded, and got this wrong in two ways.

For #211, the condition consulted the previous step's `adjoint_dependencies`, which are only fully populated during the reverse pass, so a long-range dependency (a variable reused, say, every third step) was cleared during replay and the reverse pass produced a wrong gradient. The first commit here replaced that with an identity check keeping superseded block variables alive. But that still cleared variables that remain the live block variable of their Function, relying on `saved_output` falling back to the live object — which holds the value from taping, not from the replay. As #260 shows, the adjoint of the last-use step still needs the variable as a linearisation point, so re-evaluating the functional at a new control before calling `derivative()` corrupts the gradient, which is precisely what every optimisation loop does.

The end state is a single condition covering both: a variable visited here is by construction a dependency of a block in this step, so its checkpoint is only cleared once the step's adjoint dependencies have been revised by a reverse pass and the variable is provably not among them. Before that everything is kept, which is exactly what "all adjoint dependencies are stored in memory" promises. The clear still goes through the public `checkpoint` setter, so the `is_control` guard from #257 protects control values on this path.

## Test

Against the #211 MFE the gradient matches the unscheduled tape (3205.30164... in both cases, where master gives 73765.11...). Against the #260 MFE the gradient at a new control is 187.4048 with and without the schedule, where master gives 140.5536; the step-1-last-use variant and a second evaluate/derivative cycle also agree. A control's value still survives clearing under this schedule (four timesteps of `J = sum_k m**2` at `m = 2` give `J = dJ/dm = 16`), and all 218 pyadjoint unit tests pass. Regression tests for both issues live in firedrakeproject/firedrake#5168.

### DIFF
--- a/pyadjoint/checkpointing.py
+++ b/pyadjoint/checkpointing.py
@@ -361,7 +361,7 @@ class CheckpointManager:
 
                 # Handle the case for SingleMemoryStorageSchedule
                 if isinstance(self._schedule, SingleMemoryStorageSchedule):
-                    if step > 1 and var not in self.tape.timesteps[step - 1].adjoint_dependencies:
+                    if var.output.block_variable is var:
                         var.checkpoint = None
                     continue
 
--- a/pyadjoint/checkpointing.py
+++ b/pyadjoint/checkpointing.py
@@ -361,7 +361,16 @@ class CheckpointManager:
 
                 # Handle the case for SingleMemoryStorageSchedule
                 if isinstance(self._schedule, SingleMemoryStorageSchedule):
-                    if var.output.block_variable is var:
+                    # `var` is used by a block in this step, so the adjoint of
+                    # this step may still need it as a linearisation point.
+                    # Only clear once the adjoint dependencies have been
+                    # revised by a reverse pass and `var` is provably not one
+                    # of them; before that, keeping every dependency in memory
+                    # is exactly what this schedule promises.
+                    if (
+                        current_step._revised_adj_deps
+                        and var not in current_step.adjoint_dependencies
+                    ):
                         var.checkpoint = None
                     continue
 
--- a/pyadjoint/checkpointing.py
+++ b/pyadjoint/checkpointing.py
@@ -366,7 +366,10 @@ class CheckpointManager:
                     # Only clear once the adjoint dependencies have been
                     # revised by a reverse pass and `var` is provably not one
                     # of them; before that, keeping every dependency in memory
-                    # is exactly what this schedule promises.
+                    # is exactly what this schedule promises. The trade-off is
+                    # that a forward-only recomputation retains the conservative
+                    # dependency set and so holds taping-time memory; the more
+                    # precise clearing only takes effect after the first reverse.
                     if (
                         current_step._revised_adj_deps
                         and var not in current_step.adjoint_dependencies


sghelichkhani:
I've rebased this onto master now that #257 is in, since both PRs touched the same line. The original version of this branch predated #257 and still wrote `var._checkpoint = None` directly, which would have bypassed the `is_control` guard and quietly undone #257 on the SingleMemory path. That was exactly the concern Josh raised over on firedrakeproject/firedrake#5093. The rebased commit keeps the new identity condition from this PR but clears through the public setter, so the line now reads `var.checkpoint = None`.

I've verified both behaviours locally against current firedrake main: the #211 MFE now gives an identical gradient with and without `SingleMemoryStorageSchedule` (3205.30 versus the 73765.11 you get on master), and a control's value survives clearing under the SingleMemory schedule, i.e. the firedrake #5082 scenario stays fixed on this code path too. The firedrake checkpointing tests show no change in behaviour relative to master.

sghelichkhani:
Companion test on the firedrake side is now up as firedrakeproject/firedrake#5168: a regression test for the #211 staggered-dependency pattern that fails on current master and passes with this branch, with CI temporarily pointed here.

Ig-dolci (changes requested):
The fix looks correct to me. Gating clearing on `_revised_adj_deps` is the right correctness-first approach, and using the public `checkpoint` setter preserves the `is_control` guard from #257.

The code comment already clearly explains that checkpoints are retained until a reverse pass revises the adjoint dependencies. Perhaps one minor documentation suggestion is to mention the resulting memory trade-off: forward-only recomputations retain the conservative dependency set because more precise clearing is enabled only after the first reverse.

sghelichkhani:
Added, thanks. The comment now notes the forward-only recomputation keeps the conservative set until the first reverse pass.

Ig-dolci (approved):

sghelichkhani:
@JHopeCollins @dham this has been approved since June. I checked it again against current master, including 2026.9.0: it merges cleanly, the pyadjoint tests pass, and the Firedrake adjoint tests pass, including the two regression tests in firedrakeproject/firedrake#5454, which fail without this change. Could one of you merge it when you get a chance? Once it is merged and released, I will remove the pin to this branch from #5454 so those tests run against the released pyadjoint.

JHopeCollins:
@Ig-dolci still happy with this?

PR #251, opened 2026-03-16
Fix L-BFGS-B `DeprecationWarning`: `disp` is deprecated in SciPy 1.16.0

jrmaddison:
Do not set the `disp` option by default with L-BFGS-B, as this is deprecated in SciPy 1.16.0

Also fix the logic here, which is broken by the multiline comment/string.

### DIFF
--- a/pyadjoint/optimization/optimization.py
+++ b/pyadjoint/optimization/optimization.py
@@ -62,18 +62,10 @@ def minimize_scipy_generic(rf_np, method, bounds=None, **kwargs):
     dJ = lambda m: rf_np.derivative()
     H = lambda x, p: rf_np.hessian(p)
 
-    if "options" not in kwargs:
-        kwargs["options"] = {}
-        # TODO: What to do here?
-        """
-    if rank(rf_np.rf.mpi_comm()) != 0:
-        # Shut up all processors except the first one.
-        kwargs["options"]["disp"] = False
-        """
-    else:
+    kwargs.setdefault("options", {})
+    if method not in ["L-BFGS-B"]:
         # Print out progress information by default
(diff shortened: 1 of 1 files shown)


pyadjoint/optimization/optimization.py
```diff
@@ -62,18 +62,10 @@ def minimize_scipy_generic(rf_np, method, bounds=None, **kwargs):
...
-        # Shut up all processors except the first one.
-        kwargs["options"]["disp"] = False
-        """
-    else:
+    kwargs.setdefault("options", {})
+    if method not in ["L-BFGS-B"]:
         # Print out progress information by default
-        if "disp" not in kwargs["options"]:
-            kwargs["options"]["disp"] = True
+        kwargs["options"].setdefault("disp", True)
```
JHopeCollins, line 68: Maybe we could allow passing `comm=comm` to minimize to only do the output on rank 0?
```suggestion
        if "comm" in kwargs:
            if kwargs["comm"].rank == 0:
        	    kwargs["options"].setdefault("disp", True)
        else:
            kwargs["options"].setdefault("disp", True)
```

jrmaddison: Maybe better to keep the processes synchronized?

JHopeCollins: Fair enough, I don't know what `"disp"` does well enough to say how it would interact with parallelism.

JHopeCollins (approved):

pyadjoint/optimization/optimization.py
jrmaddison: In principle I think it shouldn't matter, as all processes run the optimization, but it seems safer to have code paths match.

PR #252, opened 2026-03-19
Fix type of RFAdjointMat vector, and fix typo in RFMat.multHermitian

JHopeCollins:
The key change is on line 358: the adjoint mat should act on the _dual_ type to the functional.

The changes on lines 243-249 are just switching the names of x and y in `multHermitian` because I think it is easier to read that way: `Ax=y` for `mult` and `A^{T}y = x` for `multHermitian`.

### DIFF
--- a/pyadjoint/optimization/tao_solver.py
+++ b/pyadjoint/optimization/tao_solver.py
@@ -172,18 +172,24 @@ class RFOperation(Enum):
 
 class ReducedFunctionalMatBase:
     """
-    PETSc.Mat Python context to apply the action of a pyadjoint.ReducedFunctional.
+    Base class for PETSc.Mat Python contexts for applying the action of a ReducedFunctional.
 
-    If V is the control space and U is the functional space, each action has the following map:
-    Jhat : V -> U
-    TLM : V -> U
-    Adjoint : U* -> V*
-    Hessian : V x U* -> V* | V -> V*
+    If V is the control space and U is the functional space, then the ReducedFunctional
+    Jhat and its methods map between the following spaces:
+    * Jhat : V -> U
+    * TLM : V -> U
+    * Adjoint : U* -> V*
+    * Hessian : V x U* -> V* | V -> V*
+    Child classes implement the matrix action for a particular method of Jhat.
 
-    Child classes must implement:
-    - mult_impl
-    - multHermitian_impl
+    For the matrix action Ax=y the input x and output y will live in either
+    V, U, V*, or U* (e.g. for the tlm action x is in V and y is in U).
+
+    Child classes must implement the following (see docstrings for details):
+    - mult_impl, multHermitian_impl
     - update_adjoint
+    - x, y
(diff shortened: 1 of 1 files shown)


pyadjoint/optimization/tao_solver.py
```diff
@@ -240,13 +240,13 @@ def mult(self, A, x, y):
         if self._shift != 0:
             y.axpy(self._shift, x)
 
-    def multHermitian(self, A, x, y):
-        self.yinterface.from_petsc(x, self.x)
-        out = self.multHermitian_impl(A, self.x)
-        self.xinterface.to_petsc(y, out)
+    def multHermitian(self, AT, y, x):
+        self.yinterface.from_petsc(y, self.y)
```
jrmaddison, line 290: `self.y` is not defined. However it should be, as there's no reason for the spaces to match here, and as this PR points out the wrong space is being used.

pyadjoint/optimization/tao_solver.py
jrmaddison, line 284: `self.x`, `self.xinterface`, and `self.yinterface` are not defined in this class, and so this doesn't signal how subclasses should set these up. From the way these are used it looks like these should be abstract properties in the base class, and `x` and `y` should be cached properties in the derived classes.

pyadjoint/optimization/tao_solver.py
```diff
@@ -240,13 +240,13 @@ def mult(self, A, x, y):
...
+        out = self.multHermitian_impl(AT, self.y)
+        self.xinterface.to_petsc(x, out)
```
jrmaddison, line 292: `self.y`, `self.xinterface`, and `self.yinterface` are not defined in this class.

jrmaddison (changes requested):
I'm not sure this is quite the right fix, it looks like some refactoring is needed.

pyadjoint/optimization/tao_solver.py
jrmaddison: The alternative is to pass info into the base class constructor and have this set up the attributes.

pyadjoint/optimization/tao_solver.py
JHopeCollins: Absolutely right, that was sloppy on my part when I first refactored the base class out.

I've added abstact properties for `xinterface`, `yinterface`, `x`, and `y` in the base class, with concrete implementations for each of the TLM, Adjoint, and Hessian Mats.

JHopeCollins: @jrmaddison CI is failing, I will investigate and let you know when it's ready to review again

JHopeCollins: The failure is with a TAO test, but it's LMVM so doesn't use the ReducedFunctionalMat, and also fails on pyadjoint master. I will investigate further but it appears to be independent of this PR.

jrmaddison (approved):

PR #254, opened 2026-04-10
Truncate long block labels

connorjward:
This makes the tape much much more readable for cases where the `__str__` of a block is too large. Since this code path is only touched during visualisation I argue that losing information here is fine.


JHopeCollins (changes requested):
This is really useful, thanks! Please can we also allow passing the width through:
`Tape.visualise`->`Tape.visualise_*`->`Tape.create_graph`->`Block.create_graph`
so it can be set by the user (with the default value you already have)?

Arguably `Tape.visualise` should call `Tape.create_graph` directly and pass the graph to each of the `visualise_*` methods, which would slightly simplify passing through arguments, but that's just style.

PR #256, opened 2026-04-16
do not import `petsc4py.PETSc` until we really need it.

JHopeCollins:
When pyadjoint is imported we were checking if `petsc4py.PETSc` was available in the `TAOSolver` file whether or not the user ever actually used the `TAOSolver`.

The problem is that `import petsc4py.PETSc` will run `petsc4py.init` if it hasn't already been run, but it won't pass `sys.argv`.
This means that any PETSc options from the command line will silently be ignored. They won't even be shown by `-options_left` because PETSc never knew they even existed.

This PR delays importing `petsc4py.PETSc` and `petsctools` until they are actually needed so that the user has the chance to call `petsc4py.init` as they see fit.

### DIFF
--- a/pyadjoint/optimization/tao_solver.py
+++ b/pyadjoint/optimization/tao_solver.py
@@ -10,34 +10,6 @@ from .optimization_problem import MinimizationProblem
 from .optimization_solver import OptimizationSolver
 
 
-def import_petsctools():
-    """Return the petsctools module.
-
-    Delay importing petsctools so that we can put TAOSolver
-    in the top-level pyadjoint namespace without requiring
-    petsctools or petsc4py to be installed.
-    """
-    try:
-        import petsctools
-        return petsctools
-    except ModuleNotFoundError as err:
-        raise RuntimeError("petsctools not available") from err
-
-
-def import_petsc():
-    """Return the petsc4py.PETSc module.
-
-    Delay importing petsc4py so that we can put TAOSolver
-    in the top-level pyadjoint namespace without requiring
-    petsctools or petsc4py to be installed.
-    """
-    try:
-        from petsc4py import PETSc
-        return PETSc
-    except ModuleNotFoundError as err:
-        raise RuntimeError("PETSc not available") from err
-
-
 __all__ = [
     "TAOConvergenceError",
     "TAOSolver"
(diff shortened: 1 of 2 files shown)


JHopeCollins:
@jrmaddison this doesn't change anything with the top level `pyadjoint` namespace, but it means that `pyadjoint.optimization.tao_solver.PETSc` and `pyadjoint.optimization.tao_solver._reasons` no longer exist so I just wanted to check if you were relying on either of them being there.

pyadjoint/optimization/tao_solver.py
```diff
@@ -9,14 +10,33 @@
...
 
-try:
-    import petsc4py.PETSc as PETSc
-except ModuleNotFoundError:
-    PETSc = None
-try:
-    import petsctools
-except ModuleNotFoundError:
-    petsctools = None
+def import_petsctools():
```
connorjward: Why is this approach better than just having `from petsc4py import PETSc` or `import petsctools` inside the relevant functions?

jorgensd: Eventually one can do lazy imports:
https://peps.python.org/pep-0810/

connorjward: I'm very hesitant to use lazy imports when the imports have side effects

JHopeCollins: > Why is this approach better than just having from petsc4py import PETSc or import petsctools inside the relevant functions?

It is not, I was rushing this because it was getting in the way of something else I was trying to do so I just copied the old thing into functions. There's no need for us to do the `try` and raise our own error, we should just do as you suggest.

jrmaddison:
Seems fine to move `_tao_reasons`, ideally might be moved to petsctools. A module level `__getattr__` would also work for `PETSc`, `petsctools`, `_tao_reasons`, although as mentioned per function imports also seems fine.

pyadjoint/optimization/tao_solver.py
JHopeCollins: I've moved the imports inside each function that needs them.

JHopeCollins:
> Seems fine to move `_tao_reasons`, ideally might be moved to petsctools.

This is true, I forgot I had this old PR I didn't get around to finishing: https://github.com/firedrakeproject/petsctools/pull/12

connorjward (approved):

PR #257, opened 2026-05-11
Route checkpoint clears through the public BlockVariable setter

### LINKED ISSUE #5082: BUG: incorrect adjoint when using checkpoint scheduler
Consider the follow code:
```
from firedrake import *
from firedrake.adjoint import *
from checkpoint_schedules import SingleMemoryStorageSchedule
continue_annotation()
tape = get_working_tape()

schedule = SingleMemoryStorageSchedule()

tape.enable_checkpointing(schedule)

mesh = UnitSquareMesh(1,1)
V = FunctionSpace(mesh, "CG", 1)

m = Function(V).assign(1.0)

sumf = Function(V)
u = Function(V)
tst = TestFunction(V)
F = tst*u*dx - tst*m*m*dx

problem = NonlinearVariationalProblem(F, u)
solver = NonlinearVariationalSolver(problem)

for step in tape.timestepper(iter(range(4))):
    solver.solve()
    sumf.assign(sumf + u)

J = assemble(sumf*dx)
rf = ReducedFunctional(J, Control(m))

m0 = Function(V).assign(2.0)
print(rf(m0))
print(rf.derivative(apply_riesz=True).dat.data)
```
The taped model computes $$J=4m^2$$ and thus the derivative should be $$\frac{dJ}{dm} = 8m$$ and therefore for `m=2` I expect the derivative to be 16. With the checkpointing schedule as above however, even though the forward (re)run with `m=2` correctly produces $$J=16$$, the derivative it computes is 8 - it seems during the adjoint evaluation it reuses the old control value not the current one. If I comment out the `enable_checkpointing` line, I do get the expected result.

sghelichkhani:
The checkpoint manager clears block-variable checkpoints by writing the private attribute directly (`var._checkpoint = None`) at several places in `pyadjoint/checkpointing.py`. That bypasses the `is_control` guard on `BlockVariable.checkpoint`'s setter, so a control value installed via `Control.update` can be wiped during forward replay and the reverse pass then reads back the stale underlying value. Switching every direct-attribute write to the public setter (`var.checkpoint = None`) lets the guard do its job for controls while leaving the behaviour identical for everything else.

Closes firedrakeproject/firedrake#5082. Full diagnosis and reproducer are in that issue.

### DIFF
--- a/pyadjoint/checkpointing.py
+++ b/pyadjoint/checkpointing.py
@@ -198,17 +198,17 @@ class CheckpointManager:
                 # timestep only when required by an action from the schedule. Thus, we
                 # have to clear the checkpoint of block variables excluded from the self._global_deps.
                 for deps in deps_to_clear:
-                    deps._checkpoint = None
+                    deps.checkpoint = None
 
             self.tape.timesteps[timestep - 1].checkpoint(
                 _store_checkpointable_state, _store_adj_dependencies, self._global_deps)
             # Remove unnecessary variables in working memory from previous steps.
             for var in self.tape.timesteps[timestep - 1].checkpointable_state - self._global_deps:
-                var._checkpoint = None
+                var.checkpoint = None
 
             for block in self.tape.timesteps[timestep - 1]:
                 for out in block.get_outputs():
-                    out._checkpoint = None
+                    out.checkpoint = None
 
         if cp_action.storage == StorageType.DISK:
             # Activate disk checkpointing only in the checkpointing process.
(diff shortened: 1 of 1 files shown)


JHopeCollins:
This seems like the Right Thing conceptually. But there are failures in both the pyadjoint CI and the Firedrake CI for tests related to checkpointing. Please can you have a look at what's caused them?

EDIT: The failures on the firedrake CI are probably because that PR isn't pointed at this branch. See my comment on that PR for how to fix that.
~~The failures on the pyadjoint CI are different than the ones on the Firedrake CI though so that still needs addressing.~~

EDIT 2: The failures on the pyadjoint CI are obviously the ones you are fixing there, but the pyadjoint runs on the Docker container for the latest Firedrake `main` so of course they fail here. Ignore that comment. I shouldn't try and review code before coffee!

dham (approved):

JHopeCollins:
Thanks @sghelichkhani !

the CI failures are because of the Firedrake tests which rely on the incorrect behaviour so we can ignore them and merge.
A later versioned release will be coordinated with the next Firedrake point release.

PR #262, opened 2026-07-21
Riesz map pc

bj-chang:
This adds RieszMapPC, a preconditioner for the inner Krylov solve of pyadjoint's TAO Newton optimiser (tao_type nls). In these problems the gradient lives in the dual of the control space, but PETSc by default treats it as if it lived in the control space itself, and that mismatch makes the inner solve take more iterations as the mesh gets finer. RieszMapPC fixes this by applying the Riesz map (the correct dual-to-primal conversion, from each Control's riesz_map, e.g. "L2") as the preconditioner, which keeps the inner iteration count roughly mesh-independent, provided the Riesz map matches the norm in which the control is regularised.

### DIFF
--- a/pyadjoint/optimization/tao_solver.py
+++ b/pyadjoint/optimization/tao_solver.py
@@ -873,8 +873,6 @@ class RieszMapPC(PCBase):
         if viewer.getType() != PETSc.Viewer.Type.ASCII:
             return
 
-        viewer.printfASCII(
-            f"  Riesz map preconditioner: {type(self).__name__}\n")
+        viewer.printfASCII(f"  Riesz map preconditioner: {type(self).__name__}\n")
         for control in self.controls:
-            viewer.printfASCII(
-                f"  applying the {control.riesz_map} Riesz map\n")
+            viewer.printfASCII(f"  applying the {control.riesz_map} Riesz map\n")
--- a/pyadjoint/optimization/tao_solver.py
+++ b/pyadjoint/optimization/tao_solver.py
@@ -873,6 +873,8 @@ class RieszMapPC(PCBase):
         if viewer.getType() != PETSc.Viewer.Type.ASCII:
             return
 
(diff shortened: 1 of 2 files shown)


pyadjoint/optimization/tao_solver.py
```diff
@@ -812,3 +828,53 @@ def solve(self):
...
+    def view(self, pc, viewer=None):
+        """View object. Method usually called by PETSc with e.g. -tao_view.
+        """
+        from petsc4py import PETSc
+        if viewer is None:
+            return
+        if viewer.getType() != PETSc.Viewer.Type.ASCII:
+            return
+
+        viewer.printfASCII(
```
dham: fix indentation.

pyadjoint/optimization/tao_solver.py
```diff
@@ -9,10 +9,26 @@
...
+            raise ImportError(
+                "RieszMapPC requires petsctools, which is not installed"
+            )
+
 
 __all__ = [
     "TAOConvergenceError",
-    "TAOSolver"
+    "TAOSolver",
+    "RieszMapPC"
```
dham, line 31: import at optimization level.

bj-chang: Done by adding `from .tao_solver import RieszMapPC` in `optimization/__init__.py`  and listing `RieszMapPC` in `__all__` in that file

pyadjoint/optimization/tao_solver.py
bj-chang: I think we were talking about using `viewer.pushASCIITab()` etc previously, but I don't think that is applicable here? Unless my implementation of the view function is not right

bj-chang: Changed now to use `viewer.pushASCIITab()`

dham (approved):

PR #270, opened 2026-09-14
Fix a typo and use _ad_iadd rather than +=

jorgensd:
## Use `_ad_iadd` rather than `+=` to accumulate block variable contributions

`BlockVariable.add_adj_output` / `add_tlm_output` / `add_hessian_output` accumulated
contributions with the `+=` operator. That is wrong for any overloaded type whose `+` is
symbolic: in a UFL-based backend, `f += g` builds a `ufl.algebra.Sum` instead of adding
dof values, so the block variable stops being the type every consumer expects. Routing
accumulation through `_ad_iadd` gives backends a named hook to define real accumulation
without changing what `+` means on their public type.

### The contract change

`_ad_iadd` was in-place-only — the caller discarded its return. That cannot be
implemented for an immutable type. `AdjFloat` subclasses `float`, so the default
`OverloadedType._ad_iadd` (`self += other`) rebinds a local name and the caller's object
never changes:

```python
AdjFloat(1.0)._ad_iadd(AdjFloat(2.0))   # value stays 1.0, silently
```

So `_ad_iadd` and `_ad_imul` now **return the accumulated value**, and callers rebind.

This is backwards compatible with existing overrides: a `None` return is taken to mean
"mutated in place, keep what you had". Mutable overrides that mutate and return nothing
keep working untouched; without that guard they would reset the block variable to `None`,
after which the next contribution overwrites instead of accumulating and
`Control.get_derivative` returns zero with only a log line.

Both methods are now `@no_annotations`. They run during the reverse sweep, and
`AdjFloat.__add__` is annotated, so an unguarded `self += other` could append blocks to
the working tape.

### Why the `AdjFloat` casts (and why the `+` fallback stays)

Addressing @JHopeCollins's two review comments.

**The casts.** `AdjFloatExprBlock.evaluate_adj_component` returned
`codegen(diff=(idx,))(*inputs) * adj_input`, which is an `AdjFloat` only by accident —
when `codegen` happens to return a checkpointed input. For a constant local derivative
(`+`, `-`, any linear operator) `codegen` yields a bare `int`, and on the
`compute_derivative` path the seed is a plain `float`, so neither factor is overloaded:

```
a*a:  codegen -> AdjFloat(2.0)   adj_input -> float   out -> AdjFloat
a+a:  codegen -> int(1)          adj_input -> float   out -> float      # no _ad_iadd
```

All three `evaluate_*_component` methods now wrap their result, so an `AdjFloat` block
always hands back an `OverloadedType`. Before this, `compute_derivative` on `a + a` or
`a*a + a` raised `AttributeError: 'float' object has no attribute '_ad_iadd'`.

**The fallback.** @JHopeCollins asked whether the comment about blocks returning plain
floats is still true once the casts are added. It was not — the comment described the
behaviour the same commit removed, and it has been rewritten.

The fallback itself is still load-bearing, because the casts only fix `AdjFloatExprBlock`
and say nothing about any other block. `NumpyArraySliceBlock.evaluate_adj_component`
returns a bare `numpy.zeros(...)`, so an array sliced more than once accumulates through
the fallback:

```python
a = create_overloaded_object(np.array([-2.0, 3.0]))
ReducedFunctional(a[0] * a[0] + a[1], Control(a)).derivative()   # -> [-4., 1.]
```

Both accumulations there dispatch to the fallback on a bare `ndarray`. That case is now a
test, so the fallback is not later removed as unreachable. It uses `operator.iadd` rather
than `+`, which keeps ndarray accumulation in place and avoids an NEP 50 promotion that
would otherwise return a derivative with a different dtype than the control.

### Other callers

`ROLVector.plus` and `ROLVector.scale` discarded the return, which under the new contract
makes them silent no-ops for an immutable control — the line search would never move.
Both now rebind into `self.dat`.

`MergedConstraints.jacobian_adjoint_action` / `hessian_action` are left as they are and
commented instead: `result` is a caller-owned out-parameter, so there is no name to rebind
that the caller would see. They genuinely require a mutable workspace. This is unchanged
from `master`.

### Testing

`tests/pyadjoint/test_accumulation.py` adds 18 tests: the accumulation primitive across
immutable / mutable / `None`-returning / no-`_ad_iadd` value types, all three block
variable slots, the numpy fallback reproduction above, end-to-end derivative, TLM and
Hessian values for five expressions with repeated dependencies, and a check that
accumulation does not extend the tape.

Reverting the `_accumulate` guard alone fails three of them, so they are regression tests
rather than restatements.

Suite goes from 218 to 236 passing. Also checked downstream against `dolfinx-adjoint`,
which defines `_ad_iadd` without `__iadd__` on its UFL-backed `Function` — 130 passing,
including a new test for the case this PR exists to enable.

### Not addressed here

`BlockVariable` stores the first contribution by reference and later contributions mutate
it in place, so a block that passes its input straight through can have its own buffer
written into. This predates the PR — the old `+=` mutated in place too — and fixing it
means copying on first store or accumulating out of place, which is a separate change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pyadjoint/adjfloat.py
JHopeCollins, line 138: Should there be an `AdjFloat` cast here too?

pyadjoint/block_variable.py
```diff
@@ -1,10 +1,25 @@
...
+
+    The caller must rebind to the return value rather than relying on
+    `_ad_iadd` to mutate in place: an immutable value type cannot accumulate in
+    place, so its `_ad_iadd` returns a new object. `AdjFloat` subclasses
+    `float`, so this is the common case, not a corner one. Mutable types return
+    `self`, for which the rebinding is a no-op.
+
+    A block is not obliged to return an `OverloadedType` -- `AdjFloatExprBlock`
+    yields a plain `float` whenever the local derivative is a constant -- so
+    values without `_ad_iadd` fall back to `+`.
```
JHopeCollins: Is this still true with the `AdjFloat` casts added to the block return statements?

JHopeCollins:
FYI the Firedrake failures on the CircleCI are unrelated and we will be merging a fix into Firedrake shortly.

pyadjoint/block_variable.py
jorgensd: See PR overview:)

pyadjoint/adjfloat.py
jorgensd: See Pr overview

jorgensd:
@JHopeCollins Hopefully the PR is ready to go now:) Let me know what you think.

pyadjoint/optimization/constraints.py
```diff
@@ -88,6 +88,10 @@ def jacobian_action(self, m, dm, result):
         [c.jacobian_action(m, dm, result[i]) for (i, c) in enumerate(self.constraints)]
 
     def jacobian_adjoint_action(self, m, dp, result):
+        # `result` is an out-parameter owned by the caller, so `_ad_imul` and
+        # `_ad_iadd` must genuinely mutate it -- there is no name here to rebind
+        # that the caller would see. This restricts the constraint workspace to
+        # mutable types; an immutable one (AdjFloat) silently stays zero.
```
dham: This is in effect a new long docstring masquerading as a normal comment. We should use a proper docstring. From memory pyadjoint uses GoogleDoc conventions.

pyadjoint/adjfloat.py
```diff
@@ -103,7 +103,13 @@ def __init__(self, operator, *args, np_operator=None):
 
     def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx, prepared=None):
         adj_input, = adj_inputs
-        return self._operator.codegen(diff=(idx,))(*inputs) * adj_input
+        # Wrap so the adjoint value is always an OverloadedType. On the
+        # `compute_derivative` path the seed is a plain float (only
+        # `ReducedFunctional.derivative` runs it through
+        # `create_overloaded_object` first), and for a constant local derivative
+        # -- `+`, `-`, any linear operator -- codegen yields a bare int, so
+        # neither factor is overloaded and the product would be a plain float.
```
dham: Awful AI comment with huge run-on sentence. Please rephrase in human (possibly cutting a lot).

pyadjoint/block_variable.py
```diff
@@ -1,10 +1,36 @@
...
+
+    `_ad_iadd` returning `None` is taken to mean "mutated in place, keep what
+    you had". Overrides written against the pre-2026 contract, which documented
+    `Returns: None`, would otherwise wipe the accumulated value and silently
+    drop every contribution so far.
 
+    A block is not obliged to return an `OverloadedType` at all --
+    ``numpy_adjoint.array`` returns a bare ``ndarray`` -- so values without
+    `_ad_iadd` fall back to ``+=``, which stays in place for types that
+    implement it and rebinds for those that do not.
```
dham: another awful AI comment. Needs rewriting from scratch.

dham (changes requested):
I'm completely happy with the substantive change and the tests. The only issue is Claude's ongoing inability to write decent comments and docstrings. These need some human TLC.

pyadjoint/overloaded_type.py
```diff
@@ -219,20 +219,28 @@ def _ad_mul(self, other):
         """
         raise NotImplementedError
 
+    @no_annotations
```
JHopeCollins: This is a significant API change.
These methods being annotated is really useful for downstream libraries to be able to do arithmetic operations on any `OverloadedType` without knowing where it came from.

Looks like it was only added in the latest "Various fixes by claude (opus 5)" commit.
The PR description says:
> Both methods are now @no_annotations. They run during the reverse sweep, and
AdjFloat.__add__ is annotated, so an unguarded self += other could append blocks to
the working tape.

Isn't the fix for this to run the reverse sweep within a `stop_annotating` context manager rather than trying to guard each operation individually and unconditionally?

pyadjoint/adjfloat.py
JHopeCollins: This comment also raises the question of why the `compute_derivative` path doesn't use `create_overloaded_object` like the `ReducedFunctional.derivative` path does.
Is there a specific reason for this?

pyadjoint/overloaded_type.py
```diff
@@ -219,20 +219,28 @@ def _ad_mul(self, other):
...
+
         Args:
             other (object): The object to multiply `self` with.
                 Should at the very least accept `float` objects.
 
         Returns:
-            None
+            :obj:`OverloadedType`: `self` scaled by `other`. An override may
+                return `None` to mean "mutated in place"; callers treat that as
+                leaving the value unchanged.
```
JHopeCollins: What is the "value" in this sentence? The only other time "value" is used is referring to the return value, but it doesn't make sense for a return value of `None` to be unchanged seeing as `None` is a singleton.

pyadjoint/overloaded_type.py
jorgensd: Agreed. I've reverted this. Only added guards inside rol-solver.

pyadjoint/adjfloat.py
jorgensd: My best friend CLAUDE suggests:
>No good reason — it looks historical. `create_overloaded_object` was added to
`ReducedFunctional.derivative` in e52c2dd ("Sanitize adjoint input of
ReducedFunctional.derivative") because that method calls `_ad_mul(self.scale)`
on the seed and so needs an `OverloadedType`. `compute_derivative` only stores
the seed on the block variable, so nothing ever forced the issue there.
`compute_hessian` *does* normalise its seed (`_ad_init_zero` /
`_ad_init_object`, `drivers.py:118-123`), so `compute_derivative` is the
outlier of the three.
>
> It wouldn't remove the cast, though. The adjoint value is
`codegen(diff=(idx,))(*inputs) * adj_input`, and for a constant local
derivative (`+`, `-`, any linear operator) the SymPy-lambdified code returns a
bare Python `int`, so the product is unoverloaded whatever the seed is.
>
> And `create_overloaded_object` wouldn't be a complete fix in itself: only
`float` is registered, so `compute_derivative(J, m, adj_value=1)` would still
give a bare `int` plus a "Could not find overloaded class of type '<class
'int'>'" warning. (That warning already shows up 57 times in the current test
run, from `control.py:51`.)
>
> Happy to normalise the seed in `compute_derivative` via `_ad_init_object`,
mirroring the Hessian path, but I'd rather do it in its own PR since it changes
behaviour for downstream functional types.

JHopeCollins: > My best friend CLAUDE suggests:

But do you agree?

To clarify, I didn't mean that the change to `compute_derivative` would replace the fix here, just that the comment raised the question of whether it would also be needed/desirable.
Happy for it to be it's own PR, that makes sense.

